PR #1, opened 2023-05-16
Network and di initialization

AndreVero:
Add ApiKeyHeaderInterceptor.kt
Add NetworkModule.kt Retrofit and OkHttpClient initialization
Add TaskyApplication.kt
Add DI initialization

### DIFF
--- a/app/src/main/java/com/vero/tasky/core/di/NetworkModule.kt
+++ b/app/src/main/java/com/vero/tasky/core/di/NetworkModule.kt
@@ -7,6 +7,7 @@ import dagger.Provides
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
 import okhttp3.OkHttpClient
+import okhttp3.logging.HttpLoggingInterceptor
 import retrofit2.Retrofit
 import retrofit2.converter.moshi.MoshiConverterFactory
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/vero/tasky/core/di/NetworkModule.kt
```diff
@@ -0,0 +1,43 @@
...
+            .build()
+    }
+
+    @Provides
+    @Singleton
+    fun provideOkHttpClient(
+        apiKeyHeaderInterceptor: ApiKeyHeaderInterceptor
+    ) : OkHttpClient{
+        return OkHttpClient.Builder()
+            .addInterceptor(apiKeyHeaderInterceptor)
```
MKiperszmid: You can create the ApiKeyHeaderInterceptor directly here

.addInterceptor(ApiKeyHeaderInterceptor())

This will help you remove all the extra stuff in the interceptor that's not required (check next comment)

app/src/main/java/com/vero/tasky/core/interceptors/ApiKeyHeaderInterceptor.kt
```diff
@@ -0,0 +1,26 @@
+package com.vero.tasky.core.interceptors
+
+import com.vero.tasky.BuildConfig
+import okhttp3.Interceptor
+import okhttp3.Response
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
```
MKiperszmid: You can remove this annotation, and also the "@Inject constructor()"

app/src/main/java/com/vero/tasky/core/interceptors/ApiKeyHeaderInterceptor.kt
```diff
@@ -0,0 +1,26 @@
...
+        val newRequest = original
+            .newBuilder()
+            .addHeader(API_KEY_HEADER, BuildConfig.API_KEY)
+            .build()
+
+        return chain.proceed(request = newRequest)
+    }
+
+    companion object {
+        private const val API_KEY_HEADER = "x-api-key"
```
MKiperszmid: This is a personal opinion

If I use a string only once (like this case), I dont create a constant

Only when using it 2 or more times

app/src/main/java/com/vero/tasky/core/di/NetworkModule.kt
```diff
@@ -0,0 +1,43 @@
...
+            .readTimeout(60, TimeUnit.SECONDS)
```
MKiperszmid, line 38: You added the library of HttpLoggingInterceptor in the build.gradle, but arent using it

You should add it so you can check the network logs

MKiperszmid (approved):

PR #2, opened 2023-05-17
Add local storage, navigation and splash screen

AndreVero:
Add LocalStorageModule.kt
Add Preferences.kt
Add PreferencesImpl.kt
Add Screens.kt
Add SplashScreen.kt
Add RootNavigation.kt


app/src/main/java/com/vero/tasky/MainActivity.kt
```diff
@@ -7,20 +7,30 @@ import androidx.compose.foundation.layout.fillMaxSize
...
+import com.vero.tasky.core.presentation.components.RootNavigation
 import com.vero.tasky.ui.theme.TaskyTheme
 import dagger.hilt.android.AndroidEntryPoint
+import javax.inject.Inject
 
 @AndroidEntryPoint
 class MainActivity : ComponentActivity() {
+
+    @Inject
+    lateinit var preferences: Preferences
```
MKiperszmid: Personally I have a MainViewModel, where I handle this

app/src/main/java/com/vero/tasky/splash_feature/presentation/SplashScreen.kt
```diff
@@ -0,0 +1,45 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.res.painterResource
+import com.vero.tasky.R
+import com.vero.tasky.core.presentation.navigation.Screens
+import kotlinx.coroutines.delay
+
+@Composable
+fun SplashScreen(
```
MKiperszmid: No need for this

You can use Android's new SplashScreen library, that already does everything

Philipp has a video about it

app/src/main/java/com/vero/tasky/core/data/local/PreferencesImpl.kt
```diff
@@ -0,0 +1,28 @@
...
+class PreferencesImpl @Inject constructor(
+    application: Application
+) : Preferences {
+
+    private val pref = application.getSharedPreferences(
+        "com.vero.tasky", Context.MODE_PRIVATE)
+
+    override fun isLoggedIn() = getJwtToken() != null
+
+    override fun getJwtToken(): String? {
```
MKiperszmid: When a user logs in, you get user token, email and full name

You should store all of that (will be used in the future)

MKiperszmid (approved):

app/src/main/java/com/vero/tasky/MainActivity.kt
philipplackner: Yes, you rarely need to really inject something in an Activity other than a ViewModel. I'd keep the UI as stupid as possible and leave stuff like this up to the ViewModel which makes it also easier to test this behavior (otherwise you'd need to setup a UI test)

app/src/main/java/com/vero/tasky/MainActivity.kt
```diff
@@ -3,24 +3,41 @@ package com.vero.tasky
...
 @AndroidEntryPoint
 class MainActivity : ComponentActivity() {
+
+    private val viewModel: MainViewModel by viewModels()
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        val state = viewModel.state
+        val isLoggedIn = state.value.isLoggedIn
```
MKiperszmid:
```
val state = viewModel.state.value
val isLoggedIn = state.isLoggedIn
```

If you use state many times (not the case, but for the future) you can avoid calling the ".value" everytime by doing this



Also, you are using the "isLoggedIn" only once, so maybe you can avoid the variable and directly call it on the RootNavigation's paramenter

app/src/main/java/com/vero/tasky/MainActivity.kt
```diff
@@ -3,24 +3,41 @@ package com.vero.tasky
...
         setContent {
             TaskyTheme {
                 Surface(
                     modifier = Modifier.fillMaxSize(),
                     color = MaterialTheme.colors.background
-                ) {
-
+                ) { RootNavigation(
+                        navController = rememberNavController(),
+                        isLoggedIn = isLoggedIn
```
MKiperszmid: isLoggedIn = state.isLoggedIn

app/src/main/java/com/vero/tasky/core/data/local/PreferencesImpl.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import android.app.Application
+import android.content.Context
+import androidx.core.content.edit
+import com.vero.tasky.core.domain.local.Preferences
+import com.vero.tasky.core.domain.model.User
+import javax.inject.Inject
+
+class PreferencesImpl @Inject constructor(
+    application: Application
```
MKiperszmid: I rather you pass a type Context instead of Application
Mostly cause you will just need the context and passing the whole app doesn't add any benefit


Optional:

There are 2 ways to use Hilt. With @Binds (like you are doing), or with @Provides

Personally, I like to use the Provides one (Tho tbh, I don't see any difference between each)

In case you want to try the Provides way, you can:

Remove the "@Inject constructor". Just leave it as: 
```
class PreferenceImpl(context: Context): Preferences {
// Rest of the code
```
Create the Preference instance on the "LocalStorageModule" (I'll add a comment about it)

app/src/main/java/com/vero/tasky/core/data/local/PreferencesImpl.kt
```diff
@@ -0,0 +1,33 @@
...
+        val fullName = pref.getString(Preferences.FULL_NAME, null)
+        if (token != null && fullName != null)
+            return User(token = token, fullName = fullName)
+        return null
+    }
+
+    override fun saveUser(user: User) {
+        pref.edit {
+            putString(Preferences.TOKEN, user.token)
+            putString(Preferences.FULL_NAME, user.fullName)
```
MKiperszmid: You also have Email if I'm not mistaken

app/src/main/java/com/vero/tasky/core/di/LocalStorageModule.kt
```diff
@@ -0,0 +1,21 @@
...
+
+@Module
+@InstallIn(SingletonComponent::class)
+abstract class LocalStorageModule {
+
+    @Binds
+    @Singleton
+    abstract fun bindPreferences(
+        preferencesImpl: PreferencesImpl
+    ) : Preferences
```
MKiperszmid: Regarding my previous comment about @Binds vs @Provides

Here's how you would replace the method:
```
@Provides
@Singleton
fun providePreferences(@ApplicationContext context: Context): Preferences {
    return PreferencesImpl(context)
}

File: app/src/main/java/com/vero/tasky/core/data/local/PreferencesImpl.kt [hunk 9]
```diff
@@ -0,0 +1,33 @@
...
+import com.vero.tasky.core.domain.local.Preferences
+import com.vero.tasky.core.domain.model.User
+import javax.inject.Inject
+
+class PreferencesImpl @Inject constructor(
+    application: Application
+) : Preferences {
+
+    private val pref = application.getSharedPreferences(
+        "com.vero.tasky", Context.MODE_PRIVATE)
```
MKiperszmid: You are going to store a user's JWT token

Meaning, if someone get access to it, they can modify a user's information
You don't want that

For that, you can use an EncryptedSharedPreference

app/src/main/java/com/vero/tasky/core/domain/local/Preferences.kt
```diff
@@ -0,0 +1,17 @@
+package com.vero.tasky.core.domain.local
+
+import com.vero.tasky.core.domain.model.User
+
+interface Preferences {
```
MKiperszmid: I would name this "UserPreference"

Preferences is to vague, and just by reading the name one wouldn't understand the use of it

UserPreference is a bit more descriptive. It tells you it's a preference regarding the user

app/src/main/java/com/vero/tasky/core/di/NetworkModule.kt
```diff
@@ -40,4 +40,6 @@ object NetworkModule {
             .callTimeout(60, TimeUnit.SECONDS)
             .build()
     }
+
+
```
MKiperszmid: You added extra spaces

app/src/main/java/com/vero/tasky/core/di/LocalStorageModule.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.vero.tasky.core.domain.local.Preferences
+import dagger.Binds
+import dagger.Module
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+abstract class LocalStorageModule {
```
MKiperszmid, line 13: Also, when using provides, this doesn't need to be "abstract"

app/src/main/java/com/vero/tasky/core/presentation/components/RootNavigation.kt
```diff
@@ -0,0 +1,29 @@
...
+
+@Composable
+fun RootNavigation(
+    navController: NavHostController,
+    isLoggedIn: Boolean
+) {
+    NavHost(
+        navController = navController,
+        startDestination = if (isLoggedIn) Screens.Agenda.route else Screens.Login.route,
+        modifier = Modifier.fillMaxSize()
```
MKiperszmid: Is there a use to having this as a fillMaxSize tho?

MKiperszmid (approved):

app/src/main/java/com/vero/tasky/core/data/local/PreferencesImpl.kt
AndreVero: Email is only in a request (according to the docs), we also get userId, but I don't see the reason to save it. I will add in a future with other method

app/src/main/java/com/vero/tasky/core/data/local/PreferencesImpl.kt
AndreVero: I think about it, using @Provides will give the more clear structure? (Everything is provided in the same way) or it is just a question of preference? (Someone use binds, someone use provides - but it is better to write in one style)?

app/src/main/java/com/vero/tasky/core/presentation/components/RootNavigation.kt
AndreVero: Yeah, I thought that without this NavHost wouldn't take all possible space

app/src/main/java/com/vero/tasky/MainActivity.kt
```diff
@@ -22,11 +22,10 @@ class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        val state = viewModel.state
-        val isLoggedIn = state.value.isLoggedIn
+        val state = viewModel.state.value
 
         installSplashScreen().apply {
-            setKeepOnScreenCondition { state.value.isLoading }
+            setKeepOnScreenCondition { viewModel.state.value.isLoading }
```
AndreVero, line 28: Still need to use value here. If I take state then isLoading won't be updated

philipplackner: Not sure what the problem was, but yes this looks correct

app/src/main/java/com/vero/tasky/core/data/local/UserPreferencesImpl.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    private val masterKey = MasterKeys.getOrCreate(MasterKeys.AES256_GCM_SPEC)
+
+    private val pref = EncryptedSharedPreferences.create(
+        "com.vero.tasky1",
+        masterKey,
+        context,
+        EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV,
+        EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM
+    )
```
philipplackner, line 24: I'd probably prefer injecting these preferences instead of the context. That way you have the option to reuse the same prefs instance for different types of preferences in your app. However in case of encrypted prefs it's likely you'll only need this here, so feel free to keep this

app/src/main/java/com/vero/tasky/core/data/local/UserPreferencesImpl.kt
```diff
@@ -0,0 +1,42 @@
...
+        if (token != null && fullName != null)
+            return User(token = token, fullName = fullName)
+        return null
+    }
+
+    override fun saveUser(user: User) {
+        pref.edit {
+            putString(UserPreferences.TOKEN, user.token)
+            putString(UserPreferences.FULL_NAME, user.fullName)
+            commit()
```
philipplackner, line 39: Any reason of not using `apply()` which is the asynchronous version of `commit()`? I think AS even gives a warning about this

philipplackner (approved):

PR #3, opened 2023-05-19
Add network layer for login, register and authenticate features

AndreVero:


app/src/main/java/com/vero/tasky/auth/data/dto/UserLoginDto.kt
```diff
@@ -0,0 +1,18 @@
...
+data class UserLoginDto(
+    val token: String,
+    val userId: String,
+    val fullName: String
+)
+
+fun UserLoginDto.toUser() : User {
+    return User(
+        fullName = fullName,
+        token = token
```
MKiperszmid: Why are you not storing the userId also?

app/src/main/java/com/vero/tasky/auth/data/network/AuthApi.kt
```diff
@@ -0,0 +1,25 @@
...
+
+interface AuthApi {
+
+    @POST("register")
+    suspend fun register(
+        @Body registerRequest: RegisterRequest
+    ) : Response<Void>
+
+    @GET("authenticate")
+    suspend fun authenticate() : Response<Void>
```
MKiperszmid: I'm pretty sure you don't need to return a "Response<Void>"

app/src/main/java/com/vero/tasky/auth/data/network/AuthApi.kt
```diff
@@ -0,0 +1,25 @@
...
+import retrofit2.http.Body
+import retrofit2.http.GET
+import retrofit2.http.POST
+
+interface AuthApi {
+
+    @POST("register")
+    suspend fun register(
+        @Body registerRequest: RegisterRequest
+    ) : Response<Void>
```
MKiperszmid: Same as the comment I made on authenticate()

app/src/main/java/com/vero/tasky/auth/data/network/AuthApi.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    @GET("authenticate")
+    suspend fun authenticate() : Response<Void>
+
+    @POST("login")
+    suspend fun login(
+        @Body loginRequest: LoginRequest
+    ) : UserLoginDto
```
MKiperszmid: This is a personal opinion

But usually when I send info from the app to the backend, I call those object "Dto" ex: LoginDto
And the response objects, I call them "Response" ex: LoginResponse

app/src/main/java/com/vero/tasky/auth/data/repository/AuthRepositoryImpl.kt
```diff
@@ -0,0 +1,58 @@
...
+                RegisterRequest(
+                    fullName = fullName,
+                    password = password,
+                    email = email
+                )
+            )
+            Result.success(Unit)
+        } catch (e: Exception) {
+            e.printStackTrace()
+            Result.failure(e)
```
MKiperszmid: There's one major problem when using a simple try catch

You are also catching the exception called "CancellationException"
So if a coroutine get's canceled (user leaves the screen for example), then the coroutine won't be successfully cancelled

To fix this issue, you could:

```
return try {
        api.register(
            RegisterRequest(
                fullName = fullName,
                password = password,
                email = email
            )
        )
        Result.success(Unit)
    } catch (e: CancellationException) {
        throw e
    } catch (e: Exception) {
        e.printStackTrace()
        Result.failure(e)
}
```

In fact, to avoid having to create this try catch all the time, you can create an extension function ;)

app/src/main/java/com/vero/tasky/auth/domain/use_case/Authenticate.kt
```diff
@@ -0,0 +1,12 @@
+package com.vero.tasky.auth.domain.use_case
+
+import com.vero.tasky.auth.domain.repository.AuthRepository
+import javax.inject.Inject
+
+class Authenticate @Inject constructor(
```
MKiperszmid: I like to name my usecases as "AuthenticateUseCase"

app/src/main/java/com/vero/tasky/auth/domain/use_case/Authenticate.kt
```diff
@@ -0,0 +1,12 @@
+package com.vero.tasky.auth.domain.use_case
```
MKiperszmid: Kotlin linter isn't a big fan of having a "_" in a package name
Maybe you can rename it to "usecase" or "usecases"

MKiperszmid (approved):

app/src/main/java/com/vero/tasky/auth/data/request/LoginRequest.kt
```diff
@@ -0,0 +1,6 @@
+package com.vero.tasky.auth.data.request
```
MKiperszmid: I personally like to do the following:

data -> remote -> dto (and store the objects here)
data -> remote (store the AuthApi here)
data -> local -> entity (store the room entities here)
data -> local (store the dao and database here)

You called it "network" instead of remote. That's fine
But I would create a special "dto" package or similar to store these

MKiperszmid (approved):

PR #4, opened 2023-05-21
Add auth use cases

AndreVero:


app/src/main/java/com/vero/tasky/auth/di/MatchersModule.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.vero.tasky.auth.domain.matcher.EmailMatcher
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.android.components.ViewModelComponent
+import dagger.hilt.android.scopes.ViewModelScoped
+
+@Module
+@InstallIn(ViewModelComponent::class)
+object MatchersModule {
```
MKiperszmid, line 13: Tbh, there's not really a need to have that many modules

I personally just have 1 big "AuthModule" and put everything there

And if in the future I need to swap it into small modules, I do it
But 99% of the cases, just having 1 big one is more than enough

app/src/main/java/com/vero/tasky/auth/domain/usecase/ValidateEmailUseCase.kt
```diff
@@ -0,0 +1,14 @@
+package com.vero.tasky.auth.domain.usecase
+
+import com.vero.tasky.auth.domain.matcher.EmailMatcher
+import javax.inject.Inject
+
+class ValidateEmailUseCase @Inject constructor(
```
MKiperszmid, line 6: You can do this (@Inject constructor) or create it on a module file

I personally do it on a module file, so I can create a data class called LoginUseCases, and fill it with only the required usecases

And then just send that one class to the LoginViewModel

app/src/main/java/com/vero/tasky/auth/domain/usecase/ValidatePasswordUseCase.kt
```diff
@@ -0,0 +1,13 @@
+package com.vero.tasky.auth.domain.usecase
+
+import java.util.regex.Pattern
+
+class ValidatePasswordUseCase {
+
+    private val pattern = Pattern.compile("^(?=.*[a-z])(?=.*[A-Z])(?=.*\\d)")
+
+    operator fun invoke(password: String): Boolean {
+        return password.length > 8 && pattern.matcher(password).matches()
```
MKiperszmid, line 10: Here you could return an error if password is less than 8
And another error if password is > 8 but doesnt include a digit, lowercase or uppercase

So the user is aware of what they need to fix in their password

PR #6, opened 2023-05-22
Add LoginViewModel

AndreVero:


app/src/main/java/com/vero/tasky/auth/presentation/components/PasswordTextField.kt
```diff
@@ -24,11 +24,11 @@ fun PasswordTextField(
     hintRes: Int,
     text: String,
     onValueChange: (String) -> Unit,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    isPasswordVisible: Boolean,
+    onPasswordVisibilityUpdate: () -> Unit
 ) {
 
-    var showPassword by remember { mutableStateOf(false) }
```
MKiperszmid, line 30: I actually prefer having this state here

Even tho it's technically recommended to have all the state in one ViewModel (which you should)
In this case specifically, you will never do anything else besides changing stuff in this composable (visualTransformation and trailingIcon)

So this is one case where I'd argue that having this state here, is more beneficial than having it in the whole screen

app/src/main/java/com/vero/tasky/auth/presentation/components/PasswordTextField.kt
```diff
@@ -24,11 +24,11 @@ fun PasswordTextField(
     hintRes: Int,
     text: String,
     onValueChange: (String) -> Unit,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    isPasswordVisible: Boolean,
```
MKiperszmid, line 28: Mandatory parameters should come before optional parameters

Or else, to access it you'll have to use named parameters

MKiperszmid (approved):

PR #8, opened 2023-05-25
Add tests for login screen

AndreVero:


app/src/main/java/com/vero/tasky/auth/domain/usecase/password/ValidatePasswordUseCase.kt
```diff
@@ -4,10 +4,10 @@ import java.util.regex.Pattern
 
 class ValidatePasswordUseCase {
 
-    private val pattern = Pattern.compile("^(?=.*[a-z])(?=.*[A-Z])(?=.*\\d)")
+    private val pattern = Pattern.compile("^(?=.*[a-z])(?=.*[A-Z])(?=.*\\d).+\$")
```
MKiperszmid, line 7: I personally am not a big fan of Regex, due to this thing (You had to modify it for tests and so on)

And unless you test it 100%, you aren't fully sure what can happen

I prefer having a bunch of if statements on the code, to check specific parts

Ex:

if (password.length < 9) return Too Short
if (!password.any {it.isLowerCase()}) return No Lower Case
if (!password.any {it.isUpperCase()}) return No Upper Case

And so on

Also, if you read Tasky's requirements. The password NEEDS to have at least:
1 upper case
1 lower case
1 number

Which you aren't doing by applying the Regex



For email, on the other hand. It´s okay to use Regex (Android pattern´s Email regex). Because an email is either valid or invalid.

app/src/test/java/com/vero/tasky/auth/data/repository/AuthRepositoryFake.kt
```diff
@@ -0,0 +1,20 @@
+package com.vero.tasky.auth.data.repository
+
+import com.vero.tasky.auth.domain.repository.AuthRepository
+import com.vero.tasky.core.domain.model.User
+
+class AuthRepositoryFake : AuthRepository {
```
MKiperszmid, line 6: You should have a way to force the return to be Failure also

Because if you test viewmodels (and you send a FakeRepository to it) you need to verify it works for both Success and Failure

app/src/test/java/com/vero/tasky/auth/domain/usecase/ValidateNameUseCaseTest.kt
```diff
@@ -0,0 +1,27 @@
+package com.vero.tasky.auth.domain.usecase
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+
+class ValidateNameUseCaseTest {
+
+    private val validateNameUseCase = ValidateNameUseCase()
+
+    @Test fun `Validate name, too short name, result false`() {
```
MKiperszmid, line 10: I personally have the @Test above the function. Not on the same line
I find it easier to read

app/src/test/java/com/vero/tasky/auth/domain/usecase/password/ValidatePasswordUseCaseTest.kt
```diff
@@ -0,0 +1,30 @@
...
+
+class ValidatePasswordUseCaseTest {
+
+    private val validatePasswordUseCase = ValidatePasswordUseCase()
+
+    companion object {
+        private const val TOO_SHORT_PASSWORD = ""
+        private const val PASSWORD_WITHOUT_UPPERCASE = "12qwertyu"
+        private const val CORRECT_PASSWORD = "wertyuiWE12"
+    }
```
MKiperszmid, line 14: No need to have constants, if they are only going to be used once

app/src/test/java/com/vero/tasky/auth/domain/usecase/password/ValidatePasswordUseCaseTest.kt
```diff
@@ -0,0 +1,30 @@
+package com.vero.tasky.auth.domain.usecase.password
+
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+
+class ValidatePasswordUseCaseTest {
```
MKiperszmid, line 6: As mentioned on my previous comment

What is a password is "ABCDEFGHIJKLM"
It will return True
When according to the client´s expectations (Tasky´s usecase) it should return False

app/src/test/java/com/vero/tasky/auth/presentation/login/LoginViewModelTest.kt
```diff
@@ -0,0 +1,61 @@
...
+                validatePasswordUseCase = mockk(relaxed = true) {
+                    every { invoke(password = any()) } returns PasswordValidationResult.SUCCESS },
+                validateEmailUseCase = ValidateEmailUseCase(emailMatcher = mockk(relaxed = true)),
+                loginUseCase = LoginUseCase(AuthRepositoryFake())
+            )
+        )
+    }
+
+    @Test
+    fun `Login, correct info, return results`() = runTest {
```
MKiperszmid, line 54: What happens if the email is invalid

Or the password is invalid

Or both

Or none, but the repository returned failure?

You need to test all those cases
Best way imo , is to have an assertion on the viewmodel´s state. Since at the end of the day, that´s what you´ll be showing on the screen

MKiperszmid (approved):

app/src/main/java/com/vero/tasky/auth/domain/usecase/password/ValidatePasswordUseCase.kt
philipplackner: At least for things where you can easily write a bunch of if statements like validating a password. For complex patterns regex is totally fine and legit

app/src/test/java/com/vero/tasky/auth/domain/usecase/ValidateNameUseCaseTest.kt
```diff
@@ -0,0 +1,27 @@
...
+        val result = validateNameUseCase.invoke("")
+        assertThat(result).isFalse()
+    }
+
+    @Test fun `Validate name, too long name, result false`() {
+        val longStr = StringBuilder()
+        for (i in 0..51)
+            longStr.append(i)
+        val result = validateNameUseCase.invoke(longStr.toString())
```
philipplackner, line 19: In Kotlin there's the `buildString` function:
```
val longStr = buildString {
    repeat(52) { append(it) }
}
```

PR #10, opened 2023-05-27
Add RegistrationViewModel and UseCases

AndreVero:
Add RegistrationViewModel and UseCases, refactor Error and Email handling for Login/Registration screens, fixed loading button issue

### DIFF
--- a/app/src/main/java/com/vero/tasky/auth/presentation/login/LoginViewModel.kt
+++ b/app/src/main/java/com/vero/tasky/auth/presentation/login/LoginViewModel.kt
@@ -9,7 +9,8 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.vero.tasky.R
 import com.vero.tasky.auth.domain.usecase.LoginUseCases
-import com.vero.tasky.auth.domain.usecase.password.PasswordValidationResult
+import com.vero.tasky.auth.util.EmailErrorHandler
+import com.vero.tasky.auth.util.PasswordErrorHandler
 import com.vero.tasky.core.domain.local.UserPreferences
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.channels.Channel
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/vero/tasky/auth/presentation/login/LoginViewModel.kt
```diff
@@ -54,39 +55,34 @@ class LoginViewModel @Inject constructor(
...
-                }
+        EmailErrorHandler.validateEmail(
+            isEmailValid = state.isEmailValid,
+            onEmailIsValid = ::onEmailIsValid,
+            showError = ::showError
+        )
+    }
+
+    private fun onEmailIsValid() {
+        PasswordErrorHandler.onPasswordValid(
```
philipplackner, line 66: Maybe `PasswordErrorHandler` as a name is a bit confusing, since it has the job here to also handle correct passwords. What about `PasswordValidator` instead?

app/src/main/java/com/vero/tasky/auth/presentation/registration/RegistrationViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+    private val channel = Channel<UiRegistrationEvent>()
+    val uiEvent = channel.receiveAsFlow()
+
+    fun onEvent(event: RegistrationEvent) {
+        when(event) {
+            RegistrationEvent.OnBackPressed -> {
+                viewModelScope.launch {
+                    channel.send(UiRegistrationEvent.OnBackPressed)
+                }
+            }
```
philipplackner, line 42: You don't really need to send an event to the VM just to send it back to the UI. These events that don't change any UI state can be directly processed in the UI

app/src/main/java/com/vero/tasky/auth/presentation/registration/RegistrationViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+                    state.copy(password = event.password)
+                )
+            }
+            RegistrationEvent.SignUp -> signUp()
+        }
+    }
+
+    private fun signUp() {
+        if (state.isNameValid) {
+            EmailErrorHandler.validateEmail(
```
philipplackner, line 72: Same feedback as above

app/src/main/java/com/vero/tasky/auth/presentation/registration/RegistrationViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+                onEmailIsValid = ::onEmailIsValid,
+                showError = ::showError
+            )
+        } else {
+            showError(R.string.name_not_valid)
+        }
+    }
+
+    private fun onEmailIsValid() {
+        PasswordErrorHandler.onPasswordValid(
```
philipplackner, line 83: Function is called `onEmailIsValid` and it uses `onPasswordValid` 🤔 Mistake?

app/src/main/java/com/vero/tasky/auth/util/EmailErrorHandler.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import com.vero.tasky.R
+
+object EmailErrorHandler {
+
+    fun validateEmail(
+        isEmailValid: Boolean,
+        onEmailIsValid: () -> Unit,
+        showError: (message: Int) -> Unit
+    ) {
```
philipplackner, line 11: Instead of passing tons of callbacks, I'd recommend to return a result class instead. Something like this:
```
data class ValidationResult(
    val isValid: Boolean,
    val error: Int?
)
```

philipplackner (approved):

PR #11, opened 2023-05-28
1.0.0 registration screen

AndreVero:


app/src/main/java/com/vero/tasky/MainViewModel.kt
```diff
@@ -1,32 +1,60 @@
...
 
     private var _state = MutableStateFlow(MainState(
         isLoggedIn = preferences.isLoggedIn()
     ))
     val state = _state.asStateFlow()
 
+    var errorMessage by mutableStateOf<Int?>(null)
+        private set
+
+    fun onMessageErrorSeen() {
```
MKiperszmid, line 31: Im personally not a big fan of this for onetimeevents
I rather use Channels for onetimeevents

app/src/main/java/com/vero/tasky/auth/presentation/login/LoginViewModel.kt
```diff
@@ -54,23 +53,27 @@ class LoginViewModel @Inject constructor(
...
     }
 
-    private fun logIn() {
-        EmailErrorHandler.validateEmail(
-            isEmailValid = state.isEmailValid,
-            onEmailIsValid = ::onEmailIsValid,
-            showError = ::showError
-        )
+    private fun validateUserInfo() {
+        if (state.isEmailValid) {
```
MKiperszmid, line 57: I personally run the validations separate of each other, so if the email and password both are wrong, I show both errors
Not just email

app/src/main/java/com/vero/tasky/auth/presentation/login/LoginViewModel.kt
```diff
@@ -54,23 +53,27 @@ class LoginViewModel @Inject constructor(
...
     }
 
-    private fun onEmailIsValid() {
-        PasswordErrorHandler.onPasswordValid(
-            passwordValidationResult = loginUseCases.validatePasswordUseCase(state.password),
-            onPasswordIsValid = ::onPasswordIsValid,
-            showError = ::showError
+    private fun validatePassword() {
+        val result = PasswordValidator.validatePassword(
+            passwordValidationResult = loginUseCases.validatePasswordUseCase(state.password)
```
MKiperszmid, line 66: This could be separated into variables, to make it easier to read

val validation= loginusecase...

val result = PasswordValidator.validate(validation)

MKiperszmid, line 66: "PasswordValidator.validatePassword" is kinda redundant

Maybe name it:
"PasswordValidator.validate"

app/src/main/java/com/vero/tasky/MainActivity.kt
```diff
@@ -36,15 +36,23 @@ class MainActivity : ComponentActivity() {
...
-                    LocalScaffoldState provides rememberScaffoldState()
+                    LocalSnackbarHostState provides localSnackbarHostState
                 ) {
+                    val errorMessage = viewModel.errorMessage
+                    LaunchedEffect(errorMessage, localSnackbarHostState) {
+                        if (errorMessage == null)
+                            return@LaunchedEffect
+                        else {
+                            localSnackbarHostState.showSnackbar(getString(errorMessage))
+                            viewModel.onMessageErrorSeen()
```
MKiperszmid, line 49: Do you really need to "viewModel.onMessageErrorSeen()"?
Pretty sure the Snackbar already does that logic if you do it using Channel (which you do on RegistrationScreen)

app/src/main/java/com/vero/tasky/auth/util/PasswordValidator.kt
```diff
@@ -3,28 +3,26 @@ package com.vero.tasky.auth.util
 import com.vero.tasky.R
 import com.vero.tasky.auth.domain.usecase.password.PasswordValidationResult
 
-object PasswordErrorHandler {
+object PasswordValidator {
```
MKiperszmid, line 6: Mmmmm this object isnt in charge of validating the password
It just parses the error, and converts it to a string

I liked the ErrorHandler name a bit better
Or maybe
PasswordErrorMapper

app/src/main/java/com/vero/tasky/auth/util/PasswordValidator.kt
```diff
@@ -3,28 +3,26 @@ package com.vero.tasky.auth.util
...
 
-    fun onPasswordValid(
+    fun validatePassword(
```
MKiperszmid, line 8: And this method doesnt really validate the password

So id name it
.parse, .handle, or similar (depending on the name you decide to use for the class)

app/src/main/java/com/vero/tasky/auth/util/ValidationResult.kt
```diff
@@ -0,0 +1,8 @@
+package com.vero.tasky.auth.util
+
+import androidx.annotation.StringRes
+
+data class ValidationResult(
```
MKiperszmid, line 5: You could use a sealed class to make it easier

```
sealed class ValidationResult {
    object Valid: ValidationResult
    data class Invalid(@StringRes val error: Int): ValidationResult
}
```

MKiperszmid: Makes it a bit cleaner, since you dont need to check if isValid is false or errorMessage != Null

You just check for the type of ValidationResult

PR #12, opened 2023-05-28
Add Registration use case tests

AndreVero:


app/src/test/java/com/vero/tasky/auth/presentation/registration/RegistrationViewModelTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    @Test
+    fun `Register, email isn't valid, return error`() = runTest {
+        viewModel.onEvent(RegistrationEvent.OnEmailUpdated(""))
+        viewModel.onEvent(RegistrationEvent.SignUp)
+        viewModel.uiEvent.test {
+            val item = awaitItem()
+            Truth.assertThat(item).isInstanceOf(UiRegistrationEvent.ShowErrorMessage::class.java)
```
MKiperszmid, line 86: You could test if the error message is the one from the email

MKiperszmid (approved):

PR #13, opened 2023-05-29
Change parsing of password error and fixed small issues

AndreVero:


app/src/main/java/com/vero/tasky/auth/presentation/registration/RegistrationViewModel.kt
```diff
@@ -78,14 +79,12 @@ class RegistrationViewModel @Inject constructor(
...
+        val parsedResult = PasswordErrorParser.parse(validationResult)
 
-        if (result.isValid) {
-            signUp()
-        } else {
-            result.error?.let { showError(it) }
+        when(parsedResult) {
+            is PasswordParsedResult.Invalid -> showError(parsedResult.error)
+            PasswordParsedResult.Valid -> signUp()
         }
```
MKiperszmid, line 88: This signUp() shouldnt be part of the validatePassword function tbh

The validatePassword, should only validate the password

app/src/main/java/com/vero/tasky/auth/util/PasswordErrorParser.kt
```diff
@@ -3,26 +3,26 @@ package com.vero.tasky.auth.util
...
         passwordValidationResult: PasswordValidationResult,
-    ) : ValidationResult {
+    ) : PasswordParsedResult {
         return when (passwordValidationResult) {
             PasswordValidationResult.SUCCESS -> {
-                ValidationResult(isValid = true)
+                PasswordParsedResult.Valid
             }
             PasswordValidationResult.TOO_SHORT -> {
-                ValidationResult(error = R.string.password_is_too_short)
```
MKiperszmid, line 16: I actually liked the ValidationResult

Since it could be used for a Validation logic all across the app

If error, you return a message. Else, you return valid

Think about it
You can use it for password validation, but also...
If in the future you want to validate if something is correct (ex: A user exists) you can also use ValidationResult for success, or error message if doesnt exist

And for all the validation logic in the app

MKiperszmid (approved):

PR #14, opened 2023-05-30
Change parsing of password error and fixed small issues

AndreVero:


app/src/main/java/com/vero/tasky/auth/presentation/login/LoginViewModel.kt
```diff
@@ -54,37 +54,45 @@ class LoginViewModel @Inject constructor(
...
+        return if (state.isEmailValid) {
+            true
         } else {
             showError(R.string.email_not_valid)
+            false
         }
     }
 
-    private fun validatePassword() {
+    private fun isPasswordValid() : Boolean {
```
MKiperszmid, line 86: isEmailValid() isnt really checking if an email is valid. It just prints an error if the email is invalid
isPasswordValid() is checking if password is valid or not, and then printing an error

It's a bit weird how it's handled in my opinion
Mostly because the email validation is being handled when the user starts inputing their email, and the password is only handled when the user clicks the button

Either, you handle both on input, or both on button click, or both on input and button click

Personally, i'd just do it on button click


Also, from the UX standpoint, usually when an email is invalid or a password is invalid (the length, uppercase/lowercase, etc) you wouldnt really show a snackar. You would show an error message underneath the textfield

I know each app is different. But usually this is the case

So i'd go with that approach

On the other side. If there's an error on login (invalid credentials, internet connection, etc) Then a snackbar is usually used

app/src/main/java/com/vero/tasky/auth/util/PasswordErrorParser.kt
```diff
@@ -7,22 +7,22 @@ object PasswordErrorParser {
```
MKiperszmid, line 7: You have this class on "auth.util"
Personally when it's a util I share across all layers, I like to have it on the domain layer

Mostly because what happens if you want to have utils only on data or presentation? You might get confused and add it on the util package that's shared across all layers, and that shouldn't be the case

MKiperszmid (approved):

app/src/main/java/com/vero/tasky/auth/presentation/login/LoginViewModel.kt
philipplackner: Yeah it's what we call a side-effect - something the function does which you wouldn't expect from it's name. You can't use it without printing an error.

AndreVero: About this type of validation - I do it because of the green check mark and we add it only in the email and name, not in the password. And all errors validated on button click

AndreVero: Name of the methods I will change

PR #15, opened 2023-06-01
1.0.0 registration screen

AndreVero:


app/src/main/java/com/vero/tasky/auth/presentation/registration/RegistrationViewModel.kt
```diff
@@ -49,25 +51,29 @@ class RegistrationViewModel @Inject constructor(
...
             }
             RegistrationEvent.SignUp -> signUp()
         }
     }
 
     private fun signUp() {
-        val isNameValid = isNameValid()
-        val isEmailValid = isEmailValid()
-        val isPasswordValid = isPasswordValid()
+        val validationResult = isPasswordValid()
```
philipplackner, line 74: `isPasswordValid()` sounds like it returns a boolean. For functions that return a result, I'd rather use a different naming, such as `validatePassword()`

philipplackner (approved):

PR #17, opened 2023-06-04
Add network layer to agenda feature

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/remote/network/dto/EventDto.kt
```diff
@@ -0,0 +1,17 @@
...
+    val id: String,
+    val title: String,
+    val description: String,
+    val from: Long,
+    val to: Long,
+    val remindAt: Long,
+    val host: String,
+    val isUserEventCreator: Boolean,
+    val attendees: List<AttendeeDto>,
+    val photos: List<PhotoDto>
```
MKiperszmid, line 16: For photos and attendees you'll need to create DB relations

You have 2 options here
1- Do the relationship
2- (easier) Just save each as json string. It doesnt scale in the long run, but its easier and the odds of you changing it are low

app/src/main/java/com/vero/tasky/agenda/data/remote/network/dto/PhotoDto.kt
```diff
@@ -0,0 +1,9 @@
+package com.vero.tasky.agenda.data.remote.network.dto
+
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class PhotoDto (
+    val key: String,
+    val url: String,
```
MKiperszmid, line 8: When a photo is added to the event (before saving the event), the photo only has a uri (not key nor url). So you'd need to somehow differentiate between Local and Remote photos

app/src/main/java/com/vero/tasky/agenda/data/remote/repository/AgendaRepositoryImpl.kt
```diff
@@ -0,0 +1,36 @@
...
+
+class AgendaRepositoryImpl(
+    private val api: AgendaApi
+) : AgendaRepository {
+
+    override suspend fun getAgendaForDay(timezone: String, time: Long): Flow<Result<Agenda>> = flow {
+    }
+
+    override suspend fun syncAgenda(
+        deletedEventsIds: List<String>?,
```
MKiperszmid, line 19: Something I personally do on all my apps is

Never have Null lists
If a null is needed for a list, I just send it empty

Why? Avoid null errors, + you just cover it by checking if a null is not empty (for loops, while loops, etc already cover this)

app/src/main/java/com/vero/tasky/agenda/domain/model/Event.kt
```diff
@@ -0,0 +1,14 @@
+package com.vero.tasky.agenda.domain.model
+
+data class Event(
```
MKiperszmid, line 3: Reminder, Task and Event all share some variables

Maybe you can add a parent class, and inherit from that

app/src/main/java/com/vero/tasky/agenda/domain/model/Photo.kt
```diff
@@ -0,0 +1,6 @@
+package com.vero.tasky.agenda.domain.model
+
+data class Photo (
```
MKiperszmid, line 3: As mentioned on another comment, you have 2 type of photos

Local Photos (photos that werent uploaded yet) only have uris (path to where they are located on the phone?
Remote Photos (local photos that were successfuly uploaded to the server, now have a public url for anyone to access)

MKiperszmid (approved):

PR #18, opened 2023-06-05
Add local storage layer

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/local/dao/EventDao.kt
```diff
@@ -0,0 +1,22 @@
...
+
+    @Update
+    fun updateEvents(vararg events: EventEntity)
+
+    @Delete
+    fun deleteEvents(vararg events: EventEntity)
+
+    @Transaction
+    @Query("SELECT * FROM event")
+    fun loadEventsForDay(epoch: Long) : List<EventWithPhotosAndAttendees>
```
MKiperszmid, line 20: Something I recommend is returning a Flow when using Room
Example:
```
fun loadEventsForDay(epoch: Long) : List<EventWithPhotosAndAttendees>
```

Same for Tasks and Reminders

app/src/main/java/com/vero/tasky/agenda/data/local/entities/EventEntity.kt
```diff
@@ -0,0 +1,20 @@
...
+    @PrimaryKey val id: String,
+    val title: String,
+    val description: String?,
+    val remindAt: Long,
+    val time: Long,
+    val to: Long,
+    val host: String,
+    val isUserEventCreator: Boolean,
+    val attendees: List<Attendee>,
+    val networkPhotos: List<NetworkPhoto>
```
MKiperszmid, line 19: Just network photos?
What if you create an event while on airplane made? Then you'd need to save those local photos

app/src/main/java/com/vero/tasky/agenda/data/local/entities/EventEntity.kt
```diff
@@ -0,0 +1,20 @@
+package com.vero.tasky.agenda.data.local.entities
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+import com.vero.tasky.agenda.domain.model.Attendee
+import com.vero.tasky.agenda.domain.model.NetworkPhoto
+
+@Entity(tableName = "event")
+data class EventEntity(
+    @PrimaryKey val id: String,
```
MKiperszmid, line 10: Ids that aren't autogenerated (event, reminder, tasks, etc) should have the autogenerate parameter as false:

```
@PrimaryKey(autogenerate = false)
```

app/src/main/java/com/vero/tasky/agenda/data/local/entities/PhotoEntity.kt
```diff
@@ -0,0 +1,12 @@
+package com.vero.tasky.agenda.data.local.entities
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+@Entity
+data class PhotoEntity(
+    @PrimaryKey val id: Int? = null,
```
MKiperszmid, line 8: I think for photos, the id needs to be generated automatically
So you'd do the opposite as I mentioned before:
```
@PrimaryKey(autogenerate = true)
```

app/src/main/java/com/vero/tasky/agenda/domain/model/AgendaItem.kt
```diff
@@ -0,0 +1,9 @@
+package com.vero.tasky.agenda.domain.model
+
+interface AgendaItem {
```
MKiperszmid, line 3: An interface is usually used when you want the child to override methods

For overriding parameters, I personally recommend using a sealed class

Ex:
```
sealed class AgendaItem(
    val id: String,
    val title: String,
    val description: String,
    val time: LocalDateTime,
    val remindAt: LocalDateTime
) {
    data class Reminder(
    ...
    data class Task(
    ...
    data class Event(
    ...
}
```

app/src/main/java/com/vero/tasky/agenda/domain/model/NetworkPhoto.kt
```diff
@@ -1,6 +1,6 @@
 package com.vero.tasky.agenda.domain.model
 
-data class Photo (
+data class NetworkPhoto (
```
MKiperszmid, line 3: I like Photo (or AgendaPhoto) way more
Because you'd need to separate Local from Network/Remote photo

So you can have a sealed class as a parent (like with AgendaItem)

Example:
```
sealed class AgendaPhoto(val location: String) {
    data class Local(val localUri: String) : AgendaPhoto(localUri)
    data class Remote(val remoteUrl: String, val key: String) : AgendaPhoto(remoteUrl)
}
```

MKiperszmid (approved):

PR #19, opened 2023-06-06
Add local storage layer

AndreVero:


app/build.gradle
```diff
@@ -12,7 +12,7 @@ android {
 
     defaultConfig {
         applicationId "com.vero.tasky"
-        minSdk 24
+        minSdk 26
```
MKiperszmid, line 15: Why increase to 26?

app/src/main/java/com/vero/tasky/agenda/data/local/EventWithPhotosAndAttendees.kt
```diff
@@ -2,20 +2,47 @@ package com.vero.tasky.agenda.data.local
...
\ No newline at end of file
+    val localPhotos: List<LocalPhotoEntity>,
+    @Relation(
+        parentColumn = "id",
+        entityColumn = "eventId"
+    )
+    val attendees: List<AttendeeEntity>
+)
+
+fun EventWithPhotosAndAttendees.toEvent() : AgendaItem.Event {
```
MKiperszmid, line 29: Personally, I like to have mapper classes and have that logic there

Ex:
EventMapper

And there I have the .toDomain (you named it .toEvent, but it's the same), .toDto, toEntity, etc

Makes it easier to find all the mapping in one place

I usually add this folder in the "data" package, outside of local and remote

So:

data -> local -> all about Room
data -> remote -> all about Retrofit
data -> mappers -> AttendeeMapper.kt, EventMapper.kt, etc

app/src/main/java/com/vero/tasky/agenda/data/local/dao/EventDao.kt
```diff
@@ -16,7 +17,7 @@ interface EventDao {
     fun deleteEvents(vararg events: EventEntity)
 
     @Transaction
-    @Query("SELECT * FROM event")
-    fun loadEventsForDay(epoch: Long) : List<EventWithPhotosAndAttendees>
+    @Query("SELECT * FROM event WHERE DATE(:timestamp) = CURDATE()")
```
MKiperszmid, line 20: Never used CURDATE or similar
But that's interesting. Let's see if it works 💪

app/src/main/java/com/vero/tasky/agenda/data/local/entities/AttendeeEntity.kt
```diff
@@ -2,13 +2,25 @@ package com.vero.tasky.agenda.data.local.entities
...
     val email: String,
     val fullName: String,
     val eventId: String,
     val isGoing: Boolean,
     val remindAt: Long,
-)
\ No newline at end of file
+)
+
+fun AttendeeEntity.toAttendee() : Attendee {
```
MKiperszmid, line 17: AttendeeMapper (explained on another comment)

app/src/main/java/com/vero/tasky/agenda/data/remote/network/dto/AttendeeDto.kt
```diff
@@ -11,3 +12,14 @@ data class AttendeeDto (
     val isGoing: Boolean,
     val remindAt: Long,
 )
+
+fun AttendeeDto.toAttendeeEntity() : AttendeeEntity {
```
MKiperszmid, line 16: For example here

You have the attendee mapper logic here, but also on AttendeeEntity
Making it hard to find where it really is

app/src/main/java/com/vero/tasky/agenda/data/util/LocalDateTimeConverter.kt
```diff
@@ -0,0 +1,18 @@
...
+import java.time.ZoneOffset
+
+object LocalDateTimeConverter {
+
+    fun longToLocalDateTime(timestamp: Long): LocalDateTime {
+        return LocalDateTime.ofInstant(Instant.ofEpochMilli(timestamp), ZoneId.systemDefault())
+    }
+
+    fun localDateTimeToLong(localDateTime: LocalDateTime) : Long {
+        return localDateTime.toEpochSecond(ZoneOffset.UTC)
```
MKiperszmid, line 15: I wouldn't hardcode a zone offset
What happens if the user goes to another country, and sets a new event? You might get a weird interaction
I would use System default's offset

app/src/main/java/com/vero/tasky/agenda/domain/model/Agenda.kt
```diff
@@ -1,7 +1,5 @@
 package com.vero.tasky.agenda.domain.model
 
 data class Agenda(
-    val events: List<Event>,
-    val tasks: List<Task>,
-    val reminders: List<Reminder>
+    val agendaItems: List<AgendaItem>,
 )
```
MKiperszmid, line 5: Is this item really needed?
It just holds one list

MKiperszmid (approved):

PR #20, opened 2023-06-07
Add local storage layer + repository

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/local/dao/EventDao.kt
```diff
@@ -17,7 +16,11 @@ interface EventDao {
...
 
     @Transaction
-    @Query("SELECT * FROM event WHERE DATE(:timestamp) = CURDATE()")
-    fun loadEventsForDay(timestamp: Long) : Flow<List<EventWithPhotosAndAttendees>>
+    @Query("""
+        SELECT *
+        FROM event
+        WHERE DATE(:timestamp) = CURDATE()
+    """)
+    fun loadEventsForDay(timestamp: Long) : List<EventWithPhotosAndAttendees>
```
philipplackner, line 24: Might make sense to return a `Flow` here, so the UI updates are reactive

app/src/main/java/com/vero/tasky/agenda/data/mappers/AttendeeMapper.kt
```diff
@@ -0,0 +1,30 @@
+package com.vero.tasky.agenda.data.mappers
+
+import com.vero.tasky.agenda.data.local.entities.AttendeeEntity
+import com.vero.tasky.agenda.data.remote.network.dto.AttendeeDto
+import com.vero.tasky.agenda.domain.model.Attendee
+
+object AttendeeMapper {
```
philipplackner, line 7: I'm personally a friend of just making the mappers extension functions and not putting them in an `object`, that feels a bit more natural. For example:
```
fun AttendeeDto.toAttendeeEntity(): AttendeeEntity() {
    // ...
}
```

app/src/main/java/com/vero/tasky/agenda/data/remote/repository/AgendaRepositoryImpl.kt
```diff
@@ -18,25 +24,65 @@ class AgendaRepositoryImpl(
...
+                deletedTasksIds = deletedTasksIds,
+                deletedReminderIds = deletedReminderIds
             )
+        )
     }
 
-    override suspend fun getFullAgenda() =
-        ApiCallHandler.safeApiCall {
-            Agenda(emptyList())
+    override suspend fun getFullAgenda() = ApiCallHandler.safeApiCall {
```
philipplackner, line 56: Also for this `ApiCallHandler` I think the `object` is rather unnecessary and you could put the `safeApiCall` function just in a plain Kotlin file, since it's not more clear to me what's inside this object if I read its name (rather less clear 😅 )

app/src/main/java/com/vero/tasky/agenda/data/remote/repository/AgendaRepositoryImpl.kt
```diff
@@ -18,25 +24,65 @@ class AgendaRepositoryImpl(
...
+        supervisorScope {
+            agendaDto.events.map { eventDto ->
+                launch { eventDao.insertEvents(EventMapper.toEventEntity(eventDto)) }
+            }.forEach { job -> job.join() }
+            agendaDto.tasks.map { taskDto ->
+                launch { taskDao.insertTasks(TaskMapper.toTaskEntity(taskDto)) }
+            }.forEach { job -> job.join() }
+            agendaDto.reminders.map { reminderDto ->
+                launch { reminderDao.insertReminders(ReminderMapper.toReminderEntity(reminderDto)) }
+            }.forEach { job -> job.join() }
```
philipplackner, line 85: Joining these jobs should happen at the end of the function for all and not after every step. In your implementation all events are inserted in parallel, but it will resume with inserting tasks only when all events have been inserted. Instead just save the references to the jobs and then join them all at the end

philipplackner (approved):

PR #22, opened 2023-06-11
Add use cases and work manager

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/local/dao/DeletedAgendaItemDao.kt
```diff
@@ -0,0 +1,21 @@
+package com.vero.tasky.agenda.data.local.dao
+
+import androidx.room.*
+import com.vero.tasky.agenda.data.local.entities.DeletedAgendaItemEntity
+
+@Dao
+interface DeletedAgendaItemDao {
```
MKiperszmid, line 7: Personally, I have a table for each item
One for reminder
One for task
One for event

If not, one table has access to every single attribute, making it a "god table"

app/src/main/java/com/vero/tasky/agenda/data/local/entities/DeletedAgendaItemEntity.kt
```diff
@@ -0,0 +1,12 @@
+package com.vero.tasky.agenda.data.local.entities
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+import com.vero.tasky.agenda.domain.model.AgendaItemType
+
+@Entity(tableName = "deleted_item")
+data class DeletedAgendaItemEntity(
```
MKiperszmid, line 8: What if you want to also save the Updated/Created item that was updated/created during offline mode?

app/src/main/java/com/vero/tasky/agenda/data/workmanagerrunner/GetFullAgendaWorkManagerRunnerImpl.kt
```diff
@@ -0,0 +1,35 @@
...
+import androidx.work.Constraints
+import androidx.work.ExistingWorkPolicy
+import androidx.work.NetworkType
+import androidx.work.OneTimeWorkRequestBuilder
+import androidx.work.WorkManager
+import com.vero.tasky.agenda.domain.worker.GetFullAgendaWorker
+import com.vero.tasky.agenda.domain.workmanagerrunner.GetFullAgendaWorkManagerRunner
+import javax.inject.Inject
+
+class GetFullAgendaWorkManagerRunnerImpl @Inject constructor(
```
MKiperszmid, line 12: "WorkManagerRunner" can be replaced with "Worker"
It's the same, but with less wods

PR #23, opened 2023-06-11
Sync agenda before the start and add authorization token

AndreVero:


app/src/main/java/com/vero/tasky/agenda/domain/worker/SyncAgendaWorker.kt
```diff
@@ -30,10 +32,16 @@ class SyncAgendaWorker @AssistedInject constructor(
...
             deletedTaskIds = agendaItems.filter { it.type == AgendaItemType.Task }.map { it.id },
-            deletedReminderIds =agendaItems.filter { it.type == AgendaItemType.Reminder }.map { it.id },
+            deletedReminderIds = agendaItems.filter { it.type == AgendaItemType.Reminder }.map { it.id },
         )
-        return if (result.isSuccess)
+        return if (result.isSuccess) {
+            supervisorScope {
+                agendaItems.map {
+                    launch { dao.deleteAgendaItem(it) }
+                }.forEach { it.join() }
```
MKiperszmid, line 41: I'm pretty sure the ".forEach { it.join() }" can be replaced with a ".joinAll()"

PR #24, opened 2023-06-11
Add agenda view model

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/repository/AgendaRepositoryImpl.kt
```diff
@@ -69,6 +69,10 @@ class AgendaRepositoryImpl(
         saveAgendaItems(result)
     }
 
+    override suspend fun logOut() = safeApiCall {
+        api.logOut()
```
MKiperszmid, line 73: On logout, you could also delete all the tables from the local db

Mostly because if the user logs in with another account, you wouldn't want them to access the other user's calendar

app/src/main/java/com/vero/tasky/agenda/domain/repository/AgendaRepository.kt
```diff
@@ -19,4 +19,6 @@ interface AgendaRepository {
     ) : Result<Unit>
 
     suspend fun getFullAgenda() : Result<Unit>
+
+    suspend fun logOut() : Result<Unit>
```
MKiperszmid, line 23: This makes more sense to have it on the Authentication repository, since it's about the user's authentication status

app/src/main/java/com/vero/tasky/agenda/domain/util/UserNameParser.kt
```diff
@@ -0,0 +1,13 @@
...
+object UserNameParser {
+
+    fun toShortName(name: String) : String {
+        val nameParts = name.split(" ")
+        if (nameParts.size > 1)
+            return "${nameParts[0][0].uppercaseChar()}${nameParts[1][0].uppercaseChar()}"
+        else if (nameParts.isNotEmpty() && nameParts[0].length > 1)
+            return "${nameParts[0][0].uppercaseChar()}${nameParts[0][1].uppercaseChar()}"
+        return name.uppercase()
+    }
```
MKiperszmid, line 12: If name is empty, returns "" -> I would return Null or something like that

app/src/main/java/com/vero/tasky/agenda/presentation/agenda/AgendaState.kt
```diff
@@ -0,0 +1,15 @@
...
+import android.os.Parcelable
+import com.vero.tasky.agenda.domain.model.AgendaItem
+import kotlinx.parcelize.Parcelize
+import java.time.LocalDateTime
+
+@Parcelize
+data class AgendaState(
+    val userShortName: String = "",
+    val month: String = "",
+    val days: List<LocalDateTime> = emptyList(),
```
MKiperszmid, line 12: This would make more sense to just have it as List\<LocalDate\>

You won't have the need to use the "time" part of it if I'm not mistaken

app/src/main/java/com/vero/tasky/agenda/presentation/agenda/AgendaState.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import android.os.Parcelable
+import com.vero.tasky.agenda.domain.model.AgendaItem
+import kotlinx.parcelize.Parcelize
+import java.time.LocalDateTime
+
+@Parcelize
+data class AgendaState(
+    val userShortName: String = "",
+    val month: String = "",
```
MKiperszmid, line 11: Here you could have the month, rather than the String (There's definitely something like LocalDate.Month or similar)

app/src/main/java/com/vero/tasky/agenda/presentation/agenda/AgendaViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+    private fun getAgendaForChosenDay(day: LocalDateTime) {
+        val timestamp = LocalDateTimeConverter.localDateTimeToLong(day)
+        val timezone = TimeZone.getDefault().id
+
+        currentDayJob?.cancel()
+        currentDayJob = viewModelScope.launch {
+            agendaUseCases.getAgendaForDayUseCase(
+                timestamp = timestamp,
+                timezone = timezone
+            ).collect { items ->
```
MKiperszmid, line 79: You could use ".collectLatest" and that could help you avoid having to use a job

app/src/main/java/com/vero/tasky/agenda/presentation/agenda/AgendaViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    private fun getAgendaForChosenDay(day: LocalDateTime) {
+        val timestamp = LocalDateTimeConverter.localDateTimeToLong(day)
+        val timezone = TimeZone.getDefault().id
+
+        currentDayJob?.cancel()
+        currentDayJob = viewModelScope.launch {
+            agendaUseCases.getAgendaForDayUseCase(
+                timestamp = timestamp,
+                timezone = timezone
```
MKiperszmid, line 78: I wouldn't pass the timezone as a parameter

I would just do the "TimeZone.getDefault().id" in the repository itself

app/src/main/java/com/vero/tasky/agenda/presentation/agenda/AgendaViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+                updateState(state.copy(
+                    agendaItems = items
+                ))
+            }
+        }
+
+        viewModelScope.launch {
+            agendaUseCases.updateAgendaForDayUseCase(
+                timestamp = timestamp,
+                timezone = timezone
```
MKiperszmid, line 89: Same here regarding timezone

MKiperszmid (approved):

PR #25, opened 2023-06-14
Add new tables and log out system

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/local/CreatedEventWithPhotosAndAttendees.kt
```diff
@@ -0,0 +1,19 @@
+package com.vero.tasky.agenda.data.local
+
+import androidx.room.Embedded
+import androidx.room.Relation
+import com.vero.tasky.agenda.data.local.entities.*
+
+data class CreatedEventWithPhotosAndAttendees(
```
MKiperszmid, line 7: What's the difference between "CreatedEventWithPhotosAndAttendees" and "EventWithPhotosAndAttendees"?

If it's just cause Created = Events that werent sent to the api yet, then maybe rename it to "LocalEventWithPhotosAndAttendees" or something similar

app/src/main/java/com/vero/tasky/agenda/data/local/UpdatedEventWithPhotosAndAttendees.kt
```diff
@@ -0,0 +1,24 @@
+package com.vero.tasky.agenda.data.local
+
+import androidx.room.Embedded
+import androidx.room.Relation
+import com.vero.tasky.agenda.data.local.entities.*
+
+data class UpdatedEventWithPhotosAndAttendees(
```
MKiperszmid, line 7: And what's the use of this one?

app/src/main/java/com/vero/tasky/agenda/data/local/entities/CreatedEventEntity.kt
```diff
@@ -0,0 +1,15 @@
+package com.vero.tasky.agenda.data.local.entities
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+@Entity(tableName = "created_event")
+data class CreatedEventEntity(
```
MKiperszmid, line 7: Difference between this and EventEntity?

app/src/main/java/com/vero/tasky/agenda/data/local/entities/CreatedReminderEntity.kt
```diff
@@ -0,0 +1,14 @@
+package com.vero.tasky.agenda.data.local.entities
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+@Entity(tableName = "created_reminder")
+data class CreatedReminderEntity(
```
MKiperszmid, line 6: What's the use of this?

app/src/main/java/com/vero/tasky/agenda/data/repository/AgendaRepositoryImpl.kt
```diff
@@ -42,11 +36,12 @@ class AgendaRepositoryImpl(
         return merge(taskFlow, eventFlow, reminderFlow)
     }
 
-    override suspend fun updateAgendaForDay(timezone: String,
-                                            timestamp: Long): Result<Unit> {
+    override suspend fun updateAgendaForDay(timestamp: Long): Result<Unit> {
         return safeApiCall {
-            val networkAgendaItems = api.getAgendaForDay(timezone, timestamp)
-            saveAgendaItems(networkAgendaItems)
+            withContext(Dispatchers.IO) {
```
MKiperszmid, line 41: If i'm not mistaken, you don't need this since Retrofit already does it

app/src/main/java/com/vero/tasky/agenda/data/repository/AgendaRepositoryImpl.kt
```diff
@@ -65,12 +60,10 @@ class AgendaRepositoryImpl(
     }
 
     override suspend fun getFullAgenda() = safeApiCall {
-        val result = api.getFullAgenda()
-        saveAgendaItems(result)
-    }
-
-    override suspend fun logOut() = safeApiCall {
-        api.logOut()
+        withContext(Dispatchers.IO) {
```
MKiperszmid, line 63: Same here

app/src/main/java/com/vero/tasky/agenda/domain/util/LocalDateParser.kt
```diff
@@ -1,20 +1,20 @@
 package com.vero.tasky.agenda.domain.util
```
MKiperszmid, line 1: This is only used in the presentation layer, right?
If so, it can be moved to presentation.util

app/src/main/java/com/vero/tasky/core/presentation/navigation/RootNavigation.kt
```diff
@@ -29,6 +29,7 @@ fun RootNavigation(
                 onBackPressed = { navController.popBackStack() })
         }
         composable(route = Screens.Login.route) {
+            navController.popBackStack()
```
MKiperszmid, line 32: What would happen if the start destination is Login Screen?
Wont this close the app?

MKiperszmid (approved):

app/src/main/java/com/vero/tasky/agenda/data/local/CreatedEventWithPhotosAndAttendees.kt
AndreVero: @MKiperszmid yes, exactly for this purpose, I will change the name

app/src/main/java/com/vero/tasky/agenda/data/local/UpdatedEventWithPhotosAndAttendees.kt
AndreVero: This class will contain info about offline-updated Events

app/src/main/java/com/vero/tasky/agenda/data/local/entities/CreatedEventEntity.kt
AndreVero: Do you mean I can re-use EventEntity?

app/src/main/java/com/vero/tasky/agenda/data/local/entities/CreatedReminderEntity.kt
AndreVero: To save Reminders that was created offline

app/src/main/java/com/vero/tasky/agenda/data/repository/AgendaRepositoryImpl.kt
AndreVero: Yes, I need to wrap not all of these methods, only saveAgendaItems (because Room will throw the Exception, that I'm saving info from the main thread)

app/src/main/java/com/vero/tasky/agenda/data/local/entities/CreatedEventEntity.kt
MKiperszmid: If they are the same info, I dont see why not

app/src/main/java/com/vero/tasky/agenda/data/local/UpdatedEventWithPhotosAndAttendees.kt
MKiperszmid: Gotcha

app/src/main/java/com/vero/tasky/agenda/data/repository/AgendaRepositoryImpl.kt
MKiperszmid: Gotcha!

app/src/main/java/com/vero/tasky/MainViewModel.kt
```diff
@@ -39,14 +38,35 @@ class MainViewModel @Inject constructor(
...
+                            mainUseCases.logOutUseCase()
+                                .onSuccess { logOut() }
+                                .onFailure { channel.send(R.string.failed_to_log_out) }
+                        }
+                    }
+                    LogOutEventBusEvent.UnauthorizedException -> logOut()
+                }
+
+            }
+        }
```
philipplackner, line 56: I like collecting flows this way:
```
flow.onEach { emission ->

}.launchIn(viewModelScope)
```
This needs one less indentation :)

app/src/main/java/com/vero/tasky/MainViewModel.kt
```diff
@@ -39,14 +38,35 @@ class MainViewModel @Inject constructor(
...
+    }
+
+    private fun logOut() {
+        mainUseCases.clearDatabaseUseCase()
```
philipplackner, line 60: This doesn't really read like a function, I'd rather call it `clearDatabase`

app/src/main/java/com/vero/tasky/agenda/data/local/AgendaDatabase.kt
```diff
@@ -2,21 +2,21 @@ package com.vero.tasky.agenda.data.local
...
-    LocalPhotoEntity::class, RemotePhotoEntity::class, DeletedAgendaItemEntity::class,
-    AttendeeEntity::class], version = 1)
-@TypeConverters(Converters::class)
+@Database(entities = [AttendeeEntity::class, CreatedAttendeeEntity::class, CreatedEventEntity::class,
+    CreatedLocalPhotoEntity::class, CreatedReminderEntity::class, CreatedTaskEntity::class,
+    DeletedEventEntity::class, DeletedLocalPhotoEntity::class, DeletedReminderEntity::class,
+    DeletedTaskEntity::class, EventEntity::class, ReminderEntity::class,
+    RemotePhotoEntity::class, TaskEntity::class, UpdatedAttendeeEntity::class,
+    UpdatedEventEntity::class, UpdatedLocalPhotoEntity::class, UpdatedReminderEntity::class,
+    UpdatedTaskEntity::class], version = 1)
```
philipplackner, line 14: This approach of having one separate table for each type of modification for each type of agenda item doesn't scale well. Adding one additional type of agenda item means you have to add at least 4 tables to your DB. What about just one table?
```
@Entity
data class ModifiedAgendaItem(
    @PrimaryKey(autoGenerate = false) val id: String,
    val agendaItemType: AgendaItemType, // can be EVENT, TASK or REMINDER
    val modificationType: ModificationType // can be CREATED, DELETED or UPDATED
)
```
If you now add an additional AgendaItem type, all you need to do is add an extra enum entry to `AgendaItemType` and that's it. Every change in a table will require a DB migration, so you better want to avoid this if you can.

app/src/main/java/com/vero/tasky/agenda/data/local/dao/DeletedEventsDao.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun insertEvent(vararg events: DeletedEventEntity)
+
+    @Delete
+    fun deleteEvent(vararg events: DeletedEventEntity)
+
+    @Query("""
+        SELECT *
+        FROM deleted_task
```
philipplackner, line 17: Is this a mistake? Why load it from deleted tasks if those are events?

app/src/main/java/com/vero/tasky/agenda/data/local/dao/DeletedTasksDao.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun insertTask(vararg events: DeletedTaskEntity)
+
+    @Delete
+    fun deleteTask(vararg events: DeletedTaskEntity)
+
+    @Query("""
+        SELECT *
+        FROM deleted_event
```
philipplackner, line 17: And here you load deleted events, but actually want tasks

app/src/main/java/com/vero/tasky/agenda/data/local/entities/CreatedEventEntity.kt
philipplackner: Yeah, ideally consider my feedback from above, but your approach here would save the same event data at least twice which is unnecessary. All you need is the ID reference of an event that has been locally created

app/src/main/java/com/vero/tasky/agenda/data/repository/AgendaRepositoryImpl.kt
philipplackner: Room only throws that because you didn't use suspend functions there, so Room can't switch to the IO dispatcher under the hood (since `withContext` is only for suspend functions). Simply make the functions in your DAO all suspend functions and you don't need this `withContext` (only those that don't return a Flow)

app/src/main/java/com/vero/tasky/core/presentation/navigation/RootNavigation.kt
philipplackner: Not 100% sure but I think popBackStack does nothing if you're on the last screen. `navigateUp()` would close the app

PR #26, opened 2023-06-17
1.0.0 agenda UI

AndreVero:


app/src/main/java/com/vero/tasky/agenda/domain/model/Day.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+import java.time.LocalDate
+
+@Parcelize
+data class Day(
+    val date: LocalDate,
+    val isSelected: Boolean
+) : Parcelable
```
philipplackner, line 11: This is information which is only for the presentation layer (`isSelected` also reveals that, since that info is just needed to display something), therefore this should go in presentation

app/src/main/java/com/vero/tasky/agenda/presentation/agenda/AgendaViewModel.kt
```diff
@@ -83,16 +94,33 @@ class AgendaViewModel @Inject constructor(
...
+            updateState(state.copy(currentAgendaItem = null))
+        else {
+            val currentDayTime = LocalDateTime.now()
+            var currentAgendaItem: AgendaItem? = null
+            agendaItems.forEach { agendaItem ->
+                if (currentDayTime >= agendaItem.time)
+                    currentAgendaItem = agendaItem
+                else
+                    return@forEach
+            }
```
philipplackner, line 118: I'm a friend of using all the helpful functions from the Kotlin standard lib. In this case you could simply do this:
```
val currentAgendaItem = agendaItems.find { it.time <= currentDayTime }
```

app/src/main/java/com/vero/tasky/agenda/presentation/agenda/components/AgendaComponent.kt
```diff
@@ -0,0 +1,92 @@
...
+    onEditClick: (agendaItem: AgendaItem) -> Unit,
+    onDeleteClick: (agendaItem: AgendaItem) -> Unit,
+    onCheckChanged: (agendaItem: AgendaItem.Task) -> Unit,
+    isCurrent: Boolean
+) {
+
+    val backgroundColor = when(agendaItem) {
+        is AgendaItem.Reminder -> MaterialTheme.colors.reminder
+        is AgendaItem.Task -> MaterialTheme.colors.task
+        is AgendaItem.Event -> MaterialTheme.colors.event
```
philipplackner, line 37: I'd make these extra colors independent of the theme, since the theme is rather used to define different colors for light and dark mode, but these don't differ

philipplackner (approved):

app/src/main/java/com/vero/tasky/agenda/presentation/util/AgendaTimeParser.kt
```diff
@@ -0,0 +1,29 @@
...
+import java.util.*
+
+object AgendaTimeParser {
+
+    fun getTimeLine(agendaItem: AgendaItem) : String {
+        return when (agendaItem) {
+            is AgendaItem.Event -> {
+                val from = getParsedTimeLine(dateTime = agendaItem.time)
+                val to = getParsedTimeLine(dateTime = agendaItem.to)
+                "$from - $to"
```
MKiperszmid, line 15: This will only be used on the presentation layer, so it should be moved there

app/src/main/java/com/vero/tasky/ui/theme/Type.kt
```diff
@@ -30,6 +30,16 @@ val Typography = Typography(
         fontWeight = FontWeight.Normal,
         fontSize = 16.sp
     ),
+    body2 = TextStyle(
+        fontFamily = Inter,
```
MKiperszmid, line 34: The typography object, has a parameter called "DefaultFontFamily" (or smth like that)
Where you can set the fontfamily once, and gets applied everywhere
So you dont need to manually pass it to each body

I think Philipp doesnt have a video about it, but I have this one (in spanish). Maybe it can give you an idea: https://youtu.be/mIG_KEGr7Pc

app/src/main/java/com/vero/tasky/agenda/presentation/agenda/components/CircleCheckBox.kt
```diff
@@ -0,0 +1,45 @@
...
+            .background(MaterialTheme.colors.background)
+            .padding(1.dp)
+            .clip(CircleShape)
+            .background(MaterialTheme.colors.task),
+        contentAlignment = Alignment.Center
+    ) {
+        if (agendaItem.isDone)
+            Icon(
+                imageVector = Icons.Default.Check,
+                contentDescription = "Check box",
```
MKiperszmid, line 41: I would add all strings to the strings.xml

Content descriptions also, since they are used when a user has Voice Over enabled

MKiperszmid (approved):

app/src/main/java/com/vero/tasky/agenda/presentation/util/AgendaTimeParser.kt
AndreVero: What do you mean by that? Currently, it is in the presentation/util

PR #27, opened 2023-06-18
1.0.0 agenda UI

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/local/Converters.kt
```diff
@@ -0,0 +1,39 @@
...
+        return type.value
+    }
+
+    @TypeConverter
+    fun toAgendaItemType(value: String) : AgendaItemType {
+        return when (value) {
+            AgendaItemType.TASK.value -> AgendaItemType.TASK
+            AgendaItemType.EVENT.value -> AgendaItemType.EVENT
+            AgendaItemType.REMINDER.value -> AgendaItemType.REMINDER
+            else -> AgendaItemType.EVENT
```
MKiperszmid, line 20: On enums, you don't really need to add a .value attribute, unless you are going to use that value somewhere in the UI. In that case, it needs to be localized.

If it's not for UI, and only to store on the db, then:

To convert from String to Enum, you can use AgendaItemType.valueOf(string)

And from Enum to String you use .toString()

app/src/main/java/com/vero/tasky/agenda/domain/worker/SyncAgendaWorker.kt
```diff
@@ -43,16 +41,9 @@ class SyncAgendaWorker @AssistedInject constructor(
...
-                val eventJobs = events.map {
-                    launch { deletedEventsDao.deleteEvent(it) }
-                }
-                val reminderJobs = reminders.map {
-                    launch { deletedRemindersDao.deleteReminder(it) }
-                }
-                (taskJobs + eventJobs + reminderJobs).joinAll()
+                tasks.map {
+                    launch { modifiedAgendaItemDao.deleteAgendaItem() }
+                }.joinAll()
```
MKiperszmid, line 46: Why only tasks?

app/src/main/java/com/vero/tasky/agenda/presentation/agenda/AgendaScreen.kt
```diff
@@ -31,20 +29,27 @@ import kotlinx.coroutines.launch
 @Composable
 fun AgendaScreen(
     viewModel: AgendaViewModel = hiltViewModel(),
-    onLogOut: () -> Unit,
+    openEventScreen: (isEditable: Boolean) -> Unit,
+    openTaskScreen: (isEditable: Boolean) -> Unit,
+    openReminderScreen: (isEditable: Boolean) -> Unit
```
MKiperszmid, line 34: Here you could just have 1 lambda, and send an enum that differentiates them (Ex: AgendaItemType)

```
openScreen(item: AgendaItemType, isEditable: Boolean) -> Unit
```

Also, if it's open or edit, you need to also know which item you are opening/editing. So you'd need to also pass an item's id

app/src/main/java/com/vero/tasky/core/data/local/UserPreferencesImpl.kt
```diff
@@ -38,7 +38,6 @@ class UserPreferencesImpl @Inject constructor(
             putString(UserPreferences.TOKEN, user.token)
             putString(UserPreferences.FULL_NAME, user.fullName)
             putString(UserPreferences.USER_ID, user.userId)
-            apply()
         }
```
MKiperszmid, line 41: You need a ".apply()" at the end to fully save the sharedpreference

app/src/main/java/com/vero/tasky/core/domain/util/eventbus/AuthEventBusEvent.kt
```diff
@@ -0,0 +1,7 @@
+package com.vero.tasky.core.domain.util.eventbus
+
+sealed interface AuthEventBusEvent {
+    object UnauthorizedException: AuthEventBusEvent
+    object LogOut: AuthEventBusEvent
+    object LogIn: AuthEventBusEvent
```
MKiperszmid, line 6: Why do you need a Login on your autheventbus?

AndreVero: I need LogIn to notify my MainViewModel, that current user is loggedIn. If I won't do this and the user will log out at the same session, than login screen will not be changed, because isLoggedIn filed of MainState will be false
image

AndreVero: At the first start this field will be false, and I need to change it after login

app/src/main/java/com/vero/tasky/core/data/local/UserPreferencesImpl.kt
AndreVero: if I use pref.edit, it will be called from the API by itself 
image

MKiperszmid: Oh interesting
Wasnt aware of that

app/src/main/java/com/vero/tasky/core/domain/util/eventbus/AuthEventBusEvent.kt
MKiperszmid: Mmmm

Cant you check the user is logged in, by checking if the preferences have the token + its authenticated?

AndreVero: I still need to notify my MainViewModel about this. Because the user will log-in from the LoginScreen and after that I immediately open AgendaScreen. MainViewModel don't know, that screen was changed after login

PR #29, opened 2023-06-20
Add fixes and navigation

AndreVero:


app/src/main/java/com/vero/tasky/agenda/presentation/agenda/AgendaScreen.kt
```diff
@@ -29,9 +30,8 @@ import kotlinx.coroutines.launch
 @Composable
 fun AgendaScreen(
     viewModel: AgendaViewModel = hiltViewModel(),
-    openEventScreen: (isEditable: Boolean) -> Unit,
-    openTaskScreen: (isEditable: Boolean) -> Unit,
-    openReminderScreen: (isEditable: Boolean) -> Unit
+    openAgendaItemScreen: (agendaItem: AgendaItem, isEditable: Boolean) -> Unit,
+    openNewAgendaItemScreen: (agendaItemType: AgendaItemType) -> Unit,
```
MKiperszmid, line 34: I would rename these to
onAgendaItemClick
onNewAgendaItemClick

Mostly cause the AgendaScreen shouldn't know what should happen when the item is clicked (In this case, it shouldn't know it will open a screen)

It's optional tho. In this case, nothing will change. 
But in a future app, you might have the parameter "openNewScreen". Which at first, might open a new screen. But later gets refactored to do something different, then you'd also need to remember to rename the parameter

app/src/main/java/com/vero/tasky/core/presentation/navigation/RootNavigation.kt
```diff
@@ -22,11 +27,76 @@ fun RootNavigation(
...
+                        AgendaItemType.REMINDER -> Screens.Reminder.route
+                        AgendaItemType.TASK -> Screens.Task.route
+                        else -> Screens.Event.route
+                    }
+                    navController.navigate(route)
+                }
             )
         }
+        composable(
+            route = Screens.Event.route + "?itemId={itemId}&isEditable={isEditable}",
```
MKiperszmid, line 50: These parameters (itemId and isEditable) are being reused many times

You should have a constant or similar so if the parameter name changes in the future, you can just change it in one place

MKiperszmid (approved):

PR #30, opened 2023-06-23
Add event detail api and repository

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/remote/network/api/EventApi.kt
```diff
@@ -0,0 +1,55 @@
...
+    @Multipart
+    @POST("event")
+    suspend fun createEvent(
+        @Part id: MultipartBody.Part,
+        @Part title: MultipartBody.Part,
+        @Part description: MultipartBody.Part,
+        @Part from: MultipartBody.Part,
+        @Part to: MultipartBody.Part,
+        @Part remindAt: MultipartBody.Part,
+        @Part attendeeIds: List<MultipartBody.Part>,
```
philipplackner, line 25: There should only be 2 parts for this request.
1. The JSON serialized body which represents the created event
2. The photos

app/src/main/java/com/vero/tasky/agenda/data/remote/network/api/EventApi.kt
```diff
@@ -0,0 +1,55 @@
...
+        @Part id: MultipartBody.Part,
+        @Part title: MultipartBody.Part,
+        @Part description: MultipartBody.Part,
+        @Part from: MultipartBody.Part,
+        @Part to: MultipartBody.Part,
+        @Part remindAt: MultipartBody.Part,
+        @Part isGoing: MultipartBody.Part,
+        @Part attendeeIds: List<MultipartBody.Part>,
+        @Part deletedPhotoKeys: List<MultipartBody.Part>,
+        @Part photos: List<MultipartBody.Part>
```
philipplackner, line 41: Same here

app/src/main/java/com/vero/tasky/agenda/data/repository/EventRepositoryImpl.kt
```diff
@@ -0,0 +1,162 @@
...
+        event: AgendaItem.Event,
+    ): Result<AgendaItemUploadResult> {
+        val localPhotos = event.photos.filterIsInstance<AgendaPhoto.LocalPhoto>()
+
+        saveEvent(event = event, localPhotos = localPhotos)
+
+        var skippedPhoto = 0
+
+        safeSuspendCall {
+            withContext(Dispatchers.IO) {
```
philipplackner, line 34: It's not the responsibility of the caller of the suspend function to switch to the right context, but rather of the suspend function itself.
So in this case, it's the job of Retrofit to switch, since it executes blocking calls that should be run on the IO dispatcher. If you write your own suspend function which executes blocking non-suspending calls, you should switch the dispatcher there, so the function can be safely used from any dispatcher. This concept is called main-safety (safe to be called from the main thread)

app/src/main/java/com/vero/tasky/agenda/data/util/multipart/MultipartParser.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.vero.tasky.agenda.data.util.FileCompressor
+import com.vero.tasky.agenda.data.util.LocalDateTimeConverter
+import com.vero.tasky.agenda.domain.model.AgendaPhoto
+import com.vero.tasky.agenda.domain.model.Attendee
+import okhttp3.MultipartBody
+import okhttp3.RequestBody.Companion.asRequestBody
+import java.io.File
+import java.time.LocalDateTime
+
+class MultipartParser(
```
philipplackner, line 12: Shouldn't be necessary due to my comments above

app/src/main/java/com/vero/tasky/agenda/data/repository/EventRepositoryImpl.kt
```diff
@@ -0,0 +1,162 @@
...
+                api.createEvent(
+                    id = multipartParser.getMultipartId(event.id),
+                    title = multipartParser.getMultipartTitle(event.title),
+                    description = multipartParser.getMultipartDescription(event.description),
+                    from = multipartParser.getMultipartFrom(event.time),
+                    to = multipartParser.getMultipartTo(event.to),
+                    remindAt = multipartParser.getMultipartRemindAt(event.remindAt),
+                    attendeeIds = multipartParser.getMultipartAttendeesIds(event.attendees),
+                    photos = multipartPhotos
+                )
```
philipplackner, line 46: It also makes sense to think about doing this with WorkManager, since uploading 10 photos could take a moment with unstable internet connection. WorkManager guarantees this will happen at some point while your upload would be cancelled when the user closes the app before it is finished

philipplackner (approved):

PR #31, opened 2023-06-24
Add workers

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/local/entities/DeletedPhotoEntity.kt
```diff
@@ -0,0 +1,24 @@
...
+    foreignKeys = [
+        ForeignKey(
+            entity = EventEntity::class,
+            parentColumns = arrayOf("id"),
+            childColumns = arrayOf("eventId"),
+            onDelete = ForeignKey.CASCADE
+        )
+    ]
+)
+data class DeletedPhotoEntity(
```
MKiperszmid, line 18: What's the use of this class?

You could just save the event with less photos, and that should act the same way as deleting the other photos

app/src/main/java/com/vero/tasky/agenda/data/repository/EventRepositoryImpl.kt
```diff
@@ -18,48 +21,30 @@ class EventRepositoryImpl(
     private val api: EventApi,
     private val eventDao: EventDao,
     private val modifiedAgendaItemDao: ModifiedAgendaItemDao,
-    private val multipartParser: MultipartParser
+    private val multipartParser: MultipartParser,
+    private val createEventWorkerRunner: CreateEventWorkerRunner,
```
MKiperszmid, line 25: Personally to avoid creating 2 similar workers (one for Create, oke for Update) I just used the same worker, but changed the api destination based on if it was create or update

In that way, all stuff related to it, like the multiPartParser, can be added to the worker and not to repository

app/src/main/java/com/vero/tasky/agenda/data/repository/EventRepositoryImpl.kt
```diff
@@ -70,53 +55,48 @@ class EventRepositoryImpl(
...
-    ) : Result<AgendaItemUploadResult>{
+    ): Result<AgendaItemUploadResult> {
         val localPhotos = event.photos.filterIsInstance<AgendaPhoto.LocalPhoto>()
 
-        saveEvent(event = event, localPhotos = localPhotos)
+        saveEvent(event = event, localPhotos = localPhotos, deletedPhotoKeys = deletedPhotoKeys)
 
-        var skippedPhoto = 0
+        val multipartPhotos = multipartParser.getMultipartPhotos(localPhotos)
+        val skippedPhoto = localPhotos.size - multipartPhotos.size
```
MKiperszmid, line 88: This is a number that counts how many photos were skipped, right?

So I would make it plural

skippedPhotos

MKiperszmid: Also. Is it being used?
Im not seeing it

app/src/main/java/com/vero/tasky/agenda/data/worker/CreateEventWorker.kt
```diff
@@ -0,0 +1,71 @@
...
+    @Assisted private val workerParameters: WorkerParameters,
+    private val multipartParser: MultipartParser,
+    private val modifiedAgendaItemDao: ModifiedAgendaItemDao,
+    private val eventDao: EventDao,
+    private val api: EventApi,
+    private val moshi: Moshi,
+) : CoroutineWorker(context, workerParameters) {
+
+    override suspend fun doWork(): Result {
+        val eventId = inputData.getString(UpdateEventWorkerRunner.EVENT_ID) ?: ""
```
MKiperszmid, line 36: If eventid is null, you could just return Result.success or something

No need to make it empty and continue running the worker

app/src/main/java/com/vero/tasky/agenda/data/worker/UpdateEventWorker.kt
```diff
@@ -0,0 +1,77 @@
...
+    private val multipartParser: MultipartParser,
+    private val modifiedAgendaItemDao: ModifiedAgendaItemDao,
+    private val eventDao: EventDao,
+    private val api: EventApi,
+    private val moshi: Moshi
+) : CoroutineWorker(context, workerParameters) {
+
+    override suspend fun doWork(): Result {
+        val isGoing = inputData.getBoolean(UpdateEventWorkerRunner.IS_GOING, true)
+        val eventId = inputData.getString(UpdateEventWorkerRunner.EVENT_ID) ?: ""
```
MKiperszmid, line 37: Same as previous worker

app/src/main/java/com/vero/tasky/agenda/data/worker/UpdateEventWorker.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.vero.tasky.agenda.domain.model.AgendaItemType
+import com.vero.tasky.agenda.domain.model.ModificationType
+import com.vero.tasky.agenda.domain.workmanagerrunner.UpdateEventWorkerRunner
+import com.vero.tasky.core.data.remote.safeSuspendCall
+import dagger.assisted.Assisted
+import dagger.assisted.AssistedInject
+import okhttp3.MultipartBody
+
+@HiltWorker
+class UpdateEventWorker @AssistedInject constructor(
```
MKiperszmid, line 25: Since all the logic is pretty much similar, you can have it in 1 worker, and change api call and adapter (mentioned on another comment)

app/src/main/java/com/vero/tasky/agenda/data/workmanagerrunner/CreateEventWorkerRunnerImpl.kt
```diff
@@ -0,0 +1,32 @@
+package com.vero.tasky.agenda.data.workmanagerrunner
+
+import androidx.work.*
+import com.vero.tasky.agenda.data.worker.CreateEventWorker
+import com.vero.tasky.agenda.domain.workmanagerrunner.CreateEventWorkerRunner
+import com.vero.tasky.agenda.domain.workmanagerrunner.CreateEventWorkerRunner.Companion.EVENT_ID
+
+class CreateEventWorkerRunnerImpl(
```
MKiperszmid, line 8: Even this is the basically the same, so can be only 1

app/src/main/java/com/vero/tasky/agenda/di/AgendaModule.kt
```diff
@@ -69,19 +80,43 @@ object AgendaModule {
...
             workManager = WorkManager.getInstance(context)
         )
     }
 
+    @Provides
+    @Singleton
+    fun provideCreateEventWorkerRunner(@ApplicationContext context: Context)
+            : CreateEventWorkerRunner {
+        return CreateEventWorkerRunnerImpl(
+            workManager = WorkManager.getInstance(context)
```
MKiperszmid, line 95: You can create another Provides function, that provides the workmanager

So you can just pass the instance

Mostly because you'll create another workers, so it makes sense to have the Workmanager creation in one place

app/src/main/java/com/vero/tasky/agenda/data/local/entities/DeletedPhotoEntity.kt
AndreVero: But I need to save somewhere deletedPhotosKeys to send them later in the updated request

app/src/main/java/com/vero/tasky/agenda/data/repository/EventRepositoryImpl.kt
AndreVero: I return it in a Result.success()

app/src/main/java/com/vero/tasky/agenda/data/local/entities/DeletedPhotoEntity.kt
AndreVero: If this will be triggered, I should know what images were deleted 
image

AndreVero: To update event later

app/src/main/java/com/vero/tasky/agenda/data/repository/EventRepositoryImpl.kt
AndreVero: I still need multipartParser here to count how many images will be skipped

app/src/main/java/com/vero/tasky/agenda/data/local/entities/DeletedPhotoEntity.kt
AndreVero: Here I use it 
image

PR #32, opened 2023-06-25
Add use cases

AndreVero:


app/src/main/java/com/vero/tasky/agenda/domain/usecase/event/DeleteEventUseCase.kt
```diff
@@ -0,0 +1,10 @@
+package com.vero.tasky.agenda.domain.usecase.event
+
+import com.vero.tasky.agenda.domain.repository.EventRepository
+
+class DeleteEventUseCase(
+    private val eventRepository: EventRepository
+) {
+
+    operator fun invoke(id: String) = eventRepository.getEvent(id)
```
MKiperszmid, line 9: DeleteEventUseCase, but you are calling a .getEvent?

PR #34, opened 2023-06-25
Add event details base view model

AndreVero:


app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/EventDetailsViewModel.kt
```diff
@@ -0,0 +1,226 @@
...
+import javax.inject.Inject
+
+@HiltViewModel
+class EventDetailsViewModel @Inject constructor(
+    private val eventUseCases: EventUseCases,
+    private val savedStateHandle: SavedStateHandle,
+    userPreferences: UserPreferences
+) : ViewModel() {
+
+    private val userId = userPreferences.getUser()?.userId ?: ""
```
MKiperszmid, line 34: If userId is null, you should redirect to loginscreen or something

Id should never be null

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/EventDetailsViewModel.kt
```diff
@@ -0,0 +1,226 @@
...
+            is EventDetailsEvent.FromDateTimeChanged -> changeFromDateTime(event.time)
+            is EventDetailsEvent.ReminderChanged -> changeReminder(event.reminderRange)
+            is EventDetailsEvent.ToDateTimeChanged -> changeToDateTime(event.time)
+            EventDetailsEvent.UpdateEvent -> updateAgendaItemEvent()
+            is EventDetailsEvent.DeletePhoto -> deletePhoto(event.key)
+        }
+    }
+
+    private fun deletePhoto(key: String) {
+        val deletedPhotoKeys = mutableListOf<String>()
```
MKiperszmid, line 82: This whole method could be replaced with

updateState(state.copy(deletedPhotoKeys = state.deletedPhotoKeys + key))

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/EventDetailsViewModel.kt
```diff
@@ -0,0 +1,226 @@
...
+        viewModelScope.launch {
+            eventUseCases.updateEvent(
+                event = state.agendaItem,
+                deletedPhotoKeys = state.deletedPhotoKeys,
+                isGoing = state.isGoing
+            ).onSuccess {
+                channel.send(UiEventDetailsEvent.ShowInfoToast(it.countOfSkippedMessages.toString()))
+                channel.send(UiEventDetailsEvent.OnBackClick)
+            }.onFailure {
+                channel.send(UiEventDetailsEvent.ShowErrorMessage(R.string.default_event_details_error))
```
MKiperszmid, line 98: Since you are calling the upload logic on a WorkManager... This will never return failure if im not mistaken

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/EventDetailsViewModel.kt
```diff
@@ -0,0 +1,226 @@
...
+            }
+        }
+    }
+
+    private fun changeIsGoingState() {
+        viewModelScope.launch {
```
MKiperszmid, line 104: You can call the UpdateEvent method you created on line 88
Instead of duplicating the code

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/EventDetailsViewModel.kt
```diff
@@ -0,0 +1,226 @@
...
+    private fun changeFromDateTime(time: LocalDateTime) {
+        updateEvent(state.agendaItem.copy(time = time))
+        changeReminder(state.reminderRange)
+    }
+
+    private fun changeToDateTime(time: LocalDateTime) {
+        updateEvent(state.agendaItem.copy(to = time))
+    }
+
+    private fun createEvent() {
```
MKiperszmid, line 138: Update and create can call the same method, and then have the workmanager decide if it should call create or update

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/EventDetailsViewModel.kt
```diff
@@ -0,0 +1,226 @@
...
+                            isLoading = false,
+                            isErrorEmail = true,
+                            emailLabel = R.string.wrong_email
+                        )
+                    )
+                }
+        }
+    }
+
+    private fun updateEvent(event: AgendaItem.Event) {
```
MKiperszmid, line 213: Whats the use of this method?

Sometimes you use this, some other times you use updateState

Its confusing

Also, I would have thought that "updateEvent" actually updates the event on the db and api

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/model/ReminderRange.kt
```diff
@@ -0,0 +1,9 @@
+package com.vero.tasky.agenda.presentation.agendaevent.model
+
+enum class ReminderRange(val minutes: Long) {
```
MKiperszmid, line 3: Isnt this the same for reminder and task also?
You can just use the same one

MKiperszmid (approved):

PR #35, opened 2023-06-26
Add fixes and alarm manager invocation

AndreVero:


MKiperszmid (approved):

app/src/main/java/com/vero/tasky/agenda/presentation/components/TimePickerLabelComponent.kt
```diff
@@ -0,0 +1,48 @@
...
+
+@Composable
+fun TimePickerLabelComponent(
+    label: String,
+    value: String,
+    isEditable: Boolean,
+    onEditClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    Box(modifier = modifier.width(120.dp).clickable { if (isEditable) onEditClick() }) {
```
philipplackner, line 27: I think the clickable modifier has an enabled value to enable or disable the click lambda. This would be better here since with your version you still get a ripple effect, but nothing will happen if the user is not in edit mode which could be confusing

PR #38, opened 2023-06-30
Add edit photo screen

AndreVero:


app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/components/PhotosList.kt
```diff
@@ -54,14 +56,23 @@ fun PhotoList(
                 color = MaterialTheme.colors.primary,
                 style = MaterialTheme.typography.Inter600Size20
             )
+            Spacer(modifier = Modifier.height(8.dp))
             LazyVerticalGrid(
-                columns = GridCells.Adaptive(60.dp)
+                columns = GridCells.Adaptive(68.dp)
```
MKiperszmid, line 61: Why are you using a VerticalGrid for the photos? Arent the photos displayed in a row?
image

Pretty sure it should be a LazyRow

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/components/PhotosList.kt
```diff
@@ -54,14 +56,23 @@ fun PhotoList(
...
             ) {
                 items(photos) {
-                    PhotoListItem(photo = it, onPhotoClick = onPhotoClick)
+                    Box(modifier = Modifier
+                        .padding(8.dp)
+                        .aspectRatio(1f)) {
+                        PhotoListItem(photo = it, onPhotoClick = onPhotoClick)
+                    }
                 }
                 item {
```
MKiperszmid, line 70: This item should only show if there´s less than 10 photos If im not mistaken

app/src/main/java/com/vero/tasky/agenda/presentation/editphoto/EditPhotoScreen.kt
```diff
@@ -0,0 +1,68 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.unit.dp
+import coil.compose.AsyncImage
+import com.vero.tasky.R
+import com.vero.tasky.ui.theme.Inter600Size20
+import com.vero.tasky.ui.theme.headerText
+
+@Composable
+fun EditPhotoScreen(
+    uri: String,
```
MKiperszmid, line 25: You could pass a Uri object instead of a String

And do the .path here

Mostly so you verify it´s a valid uri and not some random string

MKiperszmid: Nvm. You are using Compose Navigation. So you can´t pass a Uri as parameter

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/components/PhotosList.kt
AndreVero: Yeh, it wasn't in the requirements, just more pretty as for me to show it like that (my 5 cents to the app)

PR #40, opened 2023-06-30
1.0.0 agenda event UI

AndreVero:


app/src/main/java/com/vero/tasky/agenda/di/AgendaModule.kt
```diff
@@ -159,15 +161,17 @@ object AgendaModule {
...
     ) : EventUseCases {
         return EventUseCases(
             checkAttendee = CheckAttendeeUseCase(eventRepository),
             saveEvent = SaveEventUseCase(eventRepository),
             deleteAttendee = DeleteAttendeeUseCase(eventRepository),
             deleteEvent = DeleteEventUseCase(eventRepository),
             fetchEvent = FetchEventUseCase(eventRepository),
-            getEvent = GetEventInfoUseCase(eventRepository)
+            getEvent = GetEventInfoUseCase(eventRepository),
+            validateEmailUseCase = ValidateEmailUseCase(emailMatcher = emailMatcher)
```
MKiperszmid, line 174: Since you use the same usecase here and on authentication
Maybe core could have a di package, that creates the emailUseCase

Same with the email matcher

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/EventDetailsEvent.kt
```diff
@@ -17,7 +18,11 @@ sealed interface EventDetailsEvent {
     class ToTimeChanged(val time: LocalTime) : EventDetailsEvent
     class ReminderChanged(val reminderRange: ReminderRange) : EventDetailsEvent
     class DeletePhoto(val key: String) : EventDetailsEvent
-    class AddAttendee(val email: String) : EventDetailsEvent
+    class OnEmailUpdated(val email: String) : EventDetailsEvent
+    object AddAttendee: EventDetailsEvent
+    object HideAddAttendeeDialog: EventDetailsEvent
+    object ShowAttendeeDialog: EventDetailsEvent
```
MKiperszmid, line 24: Something I started doing is so that every event and state that´s only related to the screen (Showing a dialog, or animation) I leave it on the screen, rather than on the ViewModel

Tho no idea what @philipplackner thinks about this (Tag him on whatsapp to check his opinion also)

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/EventDetailsState.kt
```diff
@@ -28,9 +28,12 @@ data class EventDetailsState(
     val isEditableForCreator: Boolean = false,
     val isEditableForAttendee: Boolean = false,
     val isLoading: Boolean = false,
-    val isErrorEmail: Boolean = false,
     val isGoingAttendees: List<Attendee> = emptyList(),
     val isNotGoingAttendees: List<Attendee> = emptyList(),
+    val emailAddress: String = "",
+    val isEmailValid: Boolean = false,
+    val isErrorEmail: Boolean = false,
```
MKiperszmid, line 35: If the email is valid, then the error will always be false
If there's an error, then emailvalid will always be true

Why not just have 1 single boolean?

val emailError: Boolean? = null

And set it to true, if the validation returns invalid, set it to false, if the validation returns valid

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/components/AddAttendeeDialog.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.vero.tasky.core.presentation.components.VerifiableTextField
+import com.vero.tasky.ui.theme.Inter700Size20
+
+@Composable
+fun AddAttendeeDialog(
+    isLoading: Boolean,
+    onDismissClick: () -> Unit,
+    onAddEmail: () -> Unit,
+    onEmailChanged: (String) -> Unit,
+    modifier: Modifier = Modifier,
```
MKiperszmid, line 32: Modifier should be the first optional parameter
If you have it here, then to add an email, isEmailValid, etc.. you need to do it by using named parameters

app/src/main/java/com/vero/tasky/agenda/presentation/components/ReminderComponent.kt
```diff
@@ -21,46 +19,75 @@ import com.vero.tasky.agenda.presentation.model.ReminderRange
...
-            verticalAlignment = Alignment.CenterVertically,
-            horizontalArrangement = Arrangement.SpaceBetween,
-        ) {
-            Box(modifier = Modifier
-                .clip(RoundedCornerShape(5.dp))
-                .width(35.dp)
-                .height(35.dp)
-                .background(reminderBackgroundColor)
-                .clickable(enabled = isEditable) { onReminderClick() }) {
+    var isReminderDropDownAreVisible by remember {
```
MKiperszmid, line 31: Maybe rename it to: isReminderDropDownVisible  ?

app/src/main/java/com/vero/tasky/agenda/presentation/components/ReminderComponent.kt
```diff
@@ -21,46 +19,75 @@ import com.vero.tasky.agenda.presentation.model.ReminderRange
...
-                modifier = Modifier.align(Alignment.CenterEnd)
-            )
+        if (isReminderDropDownAreVisible) {
+            DropdownMenu(
+                expanded = true,
+                onDismissRequest = {
+                    isReminderDropDownAreVisible = false
+                },
+            ) {
+                ReminderRange.values().forEach { reminderRange ->
```
MKiperszmid, line 77: This is something really minor

But on each recomposition, you'll be redrawing this for loop. Since "ReminderRange.values()" will recalculate

You could have the "ReminderRange.values()" on a remember state, so it only get's calculated once

(Maybe it needs to be derivedStateOf)

app/src/main/java/com/vero/tasky/auth/domain/usecase/LoginUseCases.kt
```diff
@@ -1,6 +1,7 @@
 package com.vero.tasky.auth.domain.usecase
 
 import com.vero.tasky.auth.domain.usecase.password.ValidatePasswordUseCase
+import com.vero.tasky.core.domain.usecase.ValidateEmailUseCase
```
MKiperszmid, line 5: Why do you have the Password one inside a password package?

app/src/main/java/com/vero/tasky/core/di/CoreModule.kt
```diff
@@ -56,4 +58,10 @@ object CoreModule {
         return AuthEventBus()
     }
 
+    @Provides
+    @Singleton
+    fun provideEmailMatcher() : EmailMatcher {
+        return EmailMatcherImpl()
+    }
```
MKiperszmid, line 65: Here you could have the ValidateEmail usecase creation

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/EventDetailsEvent.kt
AndreVero: I agree with your opinion and prefer to do it like you. In this case, I add these events and a field to my state just because of this (screenshot), I need to hide dialog after success adding, and I thought it will be clear to do it using one variable (another option is to add new UiEvent and hide dialog from LaunchEffect, but I preferred one variable option)
image

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/EventDetailsState.kt
AndreVero: Not in all cases, I use this error also to show errors from the back-end. For example, email is valid, but the back-end will return "user doesn't exist". Maybe it's good to show this error in other places (like a snackbar), but I want to show this in the same way as I do on the login/registration screens.

app/src/main/java/com/vero/tasky/auth/domain/usecase/LoginUseCases.kt
AndreVero: Since the password usecase also has one more class to use PasswordValidationResult, I thought it will be better to save these classes in a separate package 
image

MKiperszmid: Ah gotcha
Makes sense

app/src/main/java/com/vero/tasky/agenda/presentation/agendaevent/EventDetailsState.kt
MKiperszmid: Ahhh
But if you want to show an error from backend, then i'd make it a string

Since backend can return different errors

PR #43, opened 2023-07-02
1.0.0 agenda event UI

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/util/FileCompressor.kt
```diff
@@ -3,23 +3,29 @@ package com.vero.tasky.agenda.data.util
...
     }
 
-    private fun compressFile(file: File): File? {
-        val bmp = BitmapFactory.decodeFile(file.absolutePath)
+    private fun compressFile(uri: Uri): File? {
+        val inputStream = context.contentResolver.openInputStream(uri)
+        val bmp = BitmapFactory.decodeStream(inputStream)
         val resultBmp = bmp.copy(bmp.config, true)
-        val resultImageFile = File(context.cacheDir, "${file.name}.jpg")
+        val resultImageFile = File(context.cacheDir, "${UUID.randomUUID()}.jpg")
```
MKiperszmid, line 28: What if the image is a png? You'd be uploading it as jpg

app/src/main/java/com/vero/tasky/agenda/data/util/FileCompressor.kt
```diff
@@ -3,23 +3,29 @@ package com.vero.tasky.agenda.data.util
...
 class FileCompressor(private val context: Context) {
 
-    fun getCompressedFile(file: File): File? {
-        return if (file.length() / BYTE_IN_KB >= KB_IN_MB) {
-            compressFile(file)
-        } else {
+    fun getCompressedFile(uri: Uri): File? {
+        val file = compressFile(uri) ?: return null
+
+        return if (file.length() / BYTE_IN_KB <= KB_IN_MB) {
```
MKiperszmid, line 16: You are checking this here, but also on MultipartParser.kt

Why both places?

app/src/main/java/com/vero/tasky/agenda/data/worker/SaveEventWorker.kt
```diff
@@ -56,21 +62,25 @@ class SaveEventWorker @AssistedInject constructor(
             return Result.failure()
         }
 
+        var multipartPhotoFiles : List<File>? = null
```
MKiperszmid, line 65: Personally I tend to never make lists null
I always make them empty

MKiperszmid: Since you tend to always run a forEach.. if its empty, they never run
So its the same as having it null, but you remove the nullability

app/src/main/java/com/vero/tasky/agenda/data/worker/SaveEventWorker.kt
```diff
@@ -106,12 +117,25 @@ class SaveEventWorker @AssistedInject constructor(
...
+        updatePhotosEntities(eventDto, agendaItem)
+    }
+
+    private suspend fun updatePhotosEntities(eventDto: EventDto, agendaItem: EventWithPhotosAndAttendees) {
+        supervisorScope {
+            val remotePhotoJobs = eventDto.photos.map {
+                launch { eventDao.insertRemotePhotoEntity(it.toPhotoEntity(agendaItem.event.id)) }
+            }
+            val localPhotoJobs = agendaItem.localPhotos.map {
+                launch { eventDao.deleteLocalPhotos(it) }
```
MKiperszmid, line 136: You will always remove all the local photos here, right?
If so, the dao could have a "deleteAllLocalPhotos", and just call it once. Instead of once on every photo

MKiperszmid (approved):

PR #46, opened 2023-07-06
Fixed small issues

AndreVero:


app/src/main/java/com/vero/tasky/core/presentation/navigation/NavigationConstants.kt
```diff
@@ -13,6 +13,14 @@ class NavigationConstants {
...
         const val REMINDER_ITEM_ID = "reminderItemId"
         const val TASK_ITEM_ID = "taskItemId"
+
+        private const val EVENT_DEEP_LINK = "$DEEP_LINK_HOST?$EVENT_ITEM_ID="
+        private const val TASK_DEEP_LINK = "$DEEP_LINK_HOST?$TASK_ITEM_ID="
+        private const val REMINDER_DEEP_LINK = "$DEEP_LINK_HOST?$REMINDER_ITEM_ID="
+
+        fun getEventDeepLink(itemId: String = EVENT_ITEM_ID) = "$EVENT_DEEP_LINK{$itemId}"
+        fun getTaskDeepLink(itemId: String = TASK_ITEM_ID) = "$TASK_DEEP_LINK{$itemId}"
+        fun getReminderDeepLink(itemId: String = REMINDER_ITEM_ID) = "$REMINDER_DEEP_LINK{$itemId}"
```
MKiperszmid, line 23: Can´t you just have 3 constants and that´s it?

MKiperszmid (approved):

app/src/main/java/com/vero/tasky/core/presentation/navigation/NavigationConstants.kt
AndreVero: Sometimes I need to change itemId, so I made it a function

MKiperszmid: I mean you are just concatenating the id
So either have 3 constants, and you always concatenate
Or have 3 functions (like you currently have) but remove the 3 contants

PR #47, opened 2023-07-08
1.0.0 task details network layer

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/local/dao/TaskDao.kt
```diff
@@ -22,6 +22,12 @@ interface TaskDao {
     """)
     fun loadTasksForDay(from: Long, to: Long) : Flow<List<TaskEntity>>
 
+    @Query("SELECT * FROM task WHERE id = :id")
+    fun loadTaskFlow(id: String) : Flow<TaskEntity?>
```
MKiperszmid, line 26: When you are just getting 1 element, 99% of the times its not necessary to have a flow
Since you wont be listening to changes

app/src/main/java/com/vero/tasky/agenda/data/repository/TaskRepositoryImpl.kt
```diff
@@ -0,0 +1,80 @@
...
+import com.vero.tasky.agenda.domain.usecase.SyncAgendaUseCase
+import com.vero.tasky.agenda.domain.workmanagerrunner.SaveTaskRunner
+import com.vero.tasky.core.data.remote.safeSuspendCall
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.mapNotNull
+
+class TaskRepositoryImpl(
+    private val api: TaskApi,
+    private val taskDao: TaskDao,
+    private val saveTaskRunner: SaveTaskRunner,
```
MKiperszmid, line 24: You dont need a worker for the Task

Since Reminder and Task get saved instantly (you are just sending a json), theres no need for a WorkManager

On events, since you are uploading a photo (which might take a while) then it makes sense to have a WorkManager there

app/src/main/java/com/vero/tasky/agenda/data/repository/TaskRepositoryImpl.kt
```diff
@@ -0,0 +1,80 @@
...
+        )
+        return Result.success(Unit)
+    }
+
+    override fun getTask(id: String): Flow<AgendaItem.Task> {
+        return taskDao.loadTaskFlow(id).mapNotNull { it?.toTask() }
+    }
+
+    override suspend fun fetchTask(id: String) {
+        safeSuspendCall {
```
MKiperszmid, line 56: No need to make an api call here. You can just get it from the db

Since you'll always have the info in your db

app/src/main/java/com/vero/tasky/agenda/data/repository/TaskRepositoryImpl.kt
```diff
@@ -0,0 +1,80 @@
...
+            api.deleteTask(taskId = task.id)
+        }.onFailure {
+            modifiedAgendaItemDao.insertAgendaItem(
+                ModifiedAgendaItemEntity(
+                    id = task.id,
+                    agendaItemType = AgendaItemType.TASK,
+                    modificationType = ModificationType.DELETED
+                )
+            )
+            syncAgendaUseCase()
```
MKiperszmid, line 76: If it fails, you are calling the syncAgenda
Which it will probably fail since you'll probably not have internet connection (thats usually what caused the fail initially)

I much rather call this on the AgendaViewModel's init block

So it called only once, and not everytime you delete something

app/src/main/java/com/vero/tasky/agenda/data/worker/SaveTaskWorker.kt
```diff
@@ -0,0 +1,64 @@
...
+import com.vero.tasky.agenda.domain.model.AgendaItemType
+import com.vero.tasky.agenda.domain.model.ModificationType
+import com.vero.tasky.agenda.domain.workmanagerrunner.SaveTaskRunner.Companion.MODIFICATION_TYPE
+import com.vero.tasky.agenda.domain.workmanagerrunner.SaveTaskRunner.Companion.TASK_ID
+import com.vero.tasky.core.data.remote.safeSuspendCall
+import dagger.assisted.Assisted
+import dagger.assisted.AssistedInject
+
+@HiltWorker
+class SaveTaskWorker @AssistedInject constructor(
```
MKiperszmid, line 21: Unnecessary worker (explained on another comment)

app/src/main/java/com/vero/tasky/agenda/data/workmanagerrunner/SaveTaskRunnerImpl.kt
```diff
@@ -0,0 +1,38 @@
+package com.vero.tasky.agenda.data.workmanagerrunner
+
+import androidx.work.*
+import com.vero.tasky.agenda.data.worker.SaveTaskWorker
+import com.vero.tasky.agenda.domain.model.ModificationType
+import com.vero.tasky.agenda.domain.workmanagerrunner.SaveTaskRunner.Companion.MODIFICATION_TYPE
+import com.vero.tasky.agenda.domain.workmanagerrunner.SaveTaskRunner
+import com.vero.tasky.agenda.domain.workmanagerrunner.SaveTaskRunner.Companion.TASK_ID
+
+class SaveTaskRunnerImpl(
```
MKiperszmid, line 10: Unnecessary (another comment)

app/src/main/java/com/vero/tasky/agenda/domain/repository/TaskRepository.kt
```diff
@@ -0,0 +1,19 @@
...
+import kotlinx.coroutines.flow.Flow
+
+interface TaskRepository {
+
+    suspend fun saveTask(
+        task: AgendaItem.Task,
+        modificationType: ModificationType
+    ) : Result<Unit>
+
+    fun getTask(id: String): Flow<AgendaItem.Task>
```
MKiperszmid, line 14: No need for flow if you arent going to observe for updates (explained on another comment)

app/src/main/java/com/vero/tasky/agenda/data/repository/TaskRepositoryImpl.kt
AndreVero: Then I will need to write an additional Worker for syncing Agenda (tasks/reminders at a start) if the user uses the app offline?

app/src/main/java/com/vero/tasky/agenda/data/repository/TaskRepositoryImpl.kt
AndreVero: Make sense, I add this, because in some cases syncAgenda on start is called after getFullAgenda. And deleted items returned to the local DB. I need to modify my syncAgenda method to then.

app/src/main/java/com/vero/tasky/agenda/data/repository/TaskRepositoryImpl.kt
MKiperszmid: Yes exactly

If a coroutine fails, then you save the item locally (task and reminders)

And then have a worker that syncs

app/src/main/java/com/vero/tasky/agenda/data/repository/TaskRepositoryImpl.kt
MKiperszmid: Then you could call getFullAgenda, and on successful, call this syncAgenda 💪

PR #50, opened 2023-07-09
Changed task updating logic and fixed small issues

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/alarmhandler/AlarmHandlerImpl.kt
```diff
@@ -36,16 +36,18 @@ class AlarmHandlerImpl(
...
         }
-        alarmManager.setExactAndAllowWhileIdle(
-            AlarmManager.RTC_WAKEUP,
-            LocalDateTimeConverter.localDateTimeToLong(data.time) * 1000,
-            PendingIntent.getBroadcast(
-                context,
-                data.itemId.hashCode(),
-                intent,
-                PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
+        if (data.time.isAfter(LocalDateTime.now())) {
```
philipplackner, line 39: What happens here, if the user saves the data in timezone A and then this code is run in timezone B? Whenever you persistently save time values or sync these with a backend, you need to take care of also adjusting the timezones. Typically, you save all times in UTC time, but would then need to convert these to the local user's timezone when displaying them

app/src/main/java/com/vero/tasky/agenda/data/worker/SyncAgendaSavedItemsWorker.kt
```diff
@@ -0,0 +1,81 @@
...
+        if (runAttemptCount >= 3)
+            return Result.failure()
+
+        val createdItems = modifiedAgendaItemDao.loadCreatedAgendaItems()
+        val createdTasks = createdItems.filter { it.agendaItemType == AgendaItemType.TASK }
+        val createdReminders = createdItems.filter { it.agendaItemType == AgendaItemType.REMINDER }
+
+        val updatedItems = modifiedAgendaItemDao.loadUpdatedAgendaItems()
+        val updatedTasks = updatedItems.filter { it.agendaItemType == AgendaItemType.TASK }
+        val updatedReminders = updatedItems.filter { it.agendaItemType == AgendaItemType.REMINDER }
```
philipplackner, line 41: You can also use `groupedBy` here:
```
val updatedItems = modifiedAgendaItemDao.loadUpdatedAgendaItems().groupBy { it.agendaItemType }
val updatedTasks = updatedItems[AgendaItemType.TASK]
val updatedEvents = updatedItems[AgendaItemType.EVENT]
val updatedReminders = updatedItems[AgendaItemType.REMINDER]
```

app/src/main/java/com/vero/tasky/agenda/data/worker/SyncAgendaSavedItemsWorker.kt
```diff
@@ -0,0 +1,81 @@
...
+
+        if (createdTasks.isEmpty() && createdReminders.isEmpty()
+            && updatedTasks.isEmpty() && updatedReminders.isEmpty()
+        )
+            Result.success()
```
philipplackner, line 46: This block doesn't do anything, since you don't return the `Result.success()`

PR #51, opened 2023-07-09
Add ReminderRepository, ReminderApi, ReminderUseCases

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/repository/ReminderRepositoryImpl.kt
```diff
@@ -0,0 +1,74 @@
...
+        }.onFailure {
+                modifiedAgendaItemDao.insertAgendaItem(
+                    ModifiedAgendaItemEntity(
+                        id = reminder.id,
+                        agendaItemType = AgendaItemType.REMINDER,
+                        modificationType = modificationType
+                    )
+                )
+            }
+        return Result.success(Unit)
```
MKiperszmid, line 51: If you are always returning a Success...
Maybe this whole function can just be a "void" and not return anything

PR #53, opened 2023-07-11
Block photo editing in offline mode, add modification function in age…

AndreVero:
…nda screen


app/src/main/java/com/vero/tasky/agenda/presentation/eventdetails/EventDetailsViewModel.kt
```diff
@@ -58,6 +61,7 @@ class EventDetailsViewModel @Inject constructor(
                 photos = emptyList(),
                 host = user.userId
             ),
+            isConnected = connectionHandler.isConnected(),
```
MKiperszmid, line 64: What if the user losses conection while they are editing the item?
They'd still be able to add/remove photos

MKiperszmid (approved):

PR #54, opened 2023-07-12
Fixed timezone issue and add offline logic for photo edit handling

AndreVero:


app/src/main/java/com/vero/tasky/agenda/data/local/entities/AttendeeEntity.kt
```diff
@@ -16,7 +16,8 @@ import androidx.room.PrimaryKey
     ]
 )
 data class AttendeeEntity(
-    @PrimaryKey(autoGenerate = false)
+    @PrimaryKey(autoGenerate = true)
+    val id: Int? = null,
```
MKiperszmid, line 20: If im not mistaken, you could set it to be 0 by default

And that will make it so the id is not nullable (Room wont allow id 0, so it will automatically increment)

app/src/main/java/com/vero/tasky/agenda/presentation/eventdetails/EventDetailsViewModel.kt
```diff
@@ -90,6 +92,15 @@ class EventDetailsViewModel @Inject constructor(
                 filterAttendees(agendaItem.attendees)
             }.launchIn(viewModelScope)
         }
+
+        viewModelScope.launch {
+            connectionHandler.observeConnectionState().collectLatest {
+                if (it == ConnectionStatus.Available)
+                    updateState(state.copy(isConnected = true))
+                else
+                    updateState(state.copy(isConnected = false))
```
MKiperszmid, line 101: This could all be replaced with:
```
val isConnected = it == ConnectionStatus.Available
updateState(state.copy(isConnected = isConnected))

MKiperszmid (approved):

philipplackner (approved):

PR #55, opened 2023-07-15
Add tests for event details screen

AndreVero:


app/src/test/java/com/vero/tasky/agenda/data/local/dao/ReminderDaoFake.kt
```diff
@@ -20,6 +20,10 @@ class ReminderDaoFake : ReminderDao {
         return  flowOf(reminderList.toList())
     }
 
+    override suspend fun loadReminder(id: String): ReminderEntity {
+        return reminderList.find { it.id == id }!!
```
MKiperszmid, line 24: Pretty sure you can use .first instead of find
Which should return a non nullable item

app/src/test/java/com/vero/tasky/agenda/data/local/dao/TaskDaoFake.kt
```diff
@@ -17,10 +17,13 @@ class TaskDaoFake : TaskDao {
     override suspend fun deleteTasks(vararg tasks: TaskEntity) {}
 
     override fun loadTasksForDay(from: Long, to: Long): Flow<List<TaskEntity>> {
-
         return  flowOf(taskList.toList())
     }
 
+    override suspend fun loadTask(id: String): TaskEntity {
+        return taskList.find { it.id == id }!!
```
MKiperszmid, line 24: Same as above

app/src/test/java/com/vero/tasky/agenda/data/local/dao/ReminderDaoFake.kt
philipplackner: And then throws a `NoElementException` if there's no such element in the list, so you again end up with a crash. Doesn't matter overall. In this case, both are absolutely fine since it's just for test cases. This will only break if there's an issue in the test code and in that case it would tell you with a failing test.

app/src/test/java/com/vero/tasky/agenda/data/repository/EventRepositoryImplTest.kt
```diff
@@ -0,0 +1,213 @@
...
+    @Test
+    fun `Save event, one image was skipped`() = runTest {
+        val result = eventRepository.saveEvent(
+            event = event,
+            isGoing = true,
+            deletedPhotoKeys = emptyList(),
+            modificationType = ModificationType.CREATED
+        )
+        Truth.assertThat(result.getOrNull()?.countOfSkippedMessages).isEqualTo(1)
+    }
```
philipplackner, line 116: Where does it get clear that one image was skipped here?

app/src/test/java/com/vero/tasky/agenda/data/repository/EventRepositoryImplTest.kt
```diff
@@ -0,0 +1,213 @@
...
+
+    @Test
+    fun `Save event, event was saved successfully`() = runTest {
+        eventRepository.saveEvent(
+            event = event,
+            isGoing = true,
+            deletedPhotoKeys = emptyList(),
+            modificationType = ModificationType.CREATED
+        )
+        Truth.assertThat(eventDao.loadAllEvents().size).isEqualTo(1)
```
philipplackner, line 126: With Truth, you can do `assertThat(eventDao.loadAllEvents()).hasSize(1)`

app/src/test/java/com/vero/tasky/agenda/data/repository/EventRepositoryImplTest.kt
```diff
@@ -0,0 +1,213 @@
...
+        eventRepository.saveEvent(
+            event = event,
+            isGoing = true,
+            deletedPhotoKeys = emptyList(),
+            modificationType = ModificationType.CREATED
+        )
+        eventRepository.deleteEvent(
+            event = event
+        )
+        Truth.assertThat(eventDao.loadAllEvents().size).isEqualTo(0)
```
philipplackner, line 145: Here, you could do `assertThat(eventDao.loadAllEvents()).isEmpty()`

app/src/test/java/com/vero/tasky/agenda/data/repository/EventRepositoryImplTest.kt
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    @Test
+    fun `Delete event, event wasn't deleted`() = runTest {
```
philipplackner, line 149: IMO a little bit confusing test name, since I would typically expect the name is `Action, expected outcome`, so the action would be deleting an event here and the expected outcome that the event wasn't deleted which doesn't make much sense to me 🤔

app/src/test/java/com/vero/tasky/agenda/presentation/util/LocalDateParserTest.kt
```diff
@@ -0,0 +1,81 @@
...
+        assertThat(result[5].date).isEqualTo(
+            LocalDate.of(
+                2023, 1, 6
+            )
+        )
+        assertThat(result[6].date).isEqualTo(
+            LocalDate.of(
+                2023, 1, 7
+            )
+        )
```
philipplackner, line 53: This could be done in a loop:
```
for(i in result.indices) {
    assertThat(result[i].date).isEqualTo(LocalDate.of(2023, 1, i + 1))
}
```

PR #58, opened 2023-07-16
Add room tests

AndreVero:


app/src/androidTest/java/com/vero/tasky/agenda/data/local/dao/EventDaoTest.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun closeDatabase() {
+        database.close()
+    }
+
+    @Test
+    fun getEventsForDay_returnSuccessfully() = runTest {
+        eventDao.insertEvents(
+            event.toEventEntity(),
+            event.copy(id = "2", time = dateTimeEndOfDay).toEventEntity(),
+            event.copy(id = "3", time = dateTimeStartOfDay.plusDays(1)).toEventEntity(),
```
MKiperszmid, line 68: You should also add a minusDays(1) to verify that one also doesnt appear in the result

PR #60, opened 2023-07-18
Add login end to end test

AndreVero:


app/build.gradle
```diff
@@ -122,6 +124,9 @@ dependencies {
     androidTestImplementation 'app.cash.turbine:turbine:0.7.0'
     androidTestImplementation "androidx.arch.core:core-testing:2.2.0"
     androidTestImplementation "androidx.work:work-testing:$work_version"
+    androidTestImplementation 'com.google.dagger:hilt-android-testing:2.44'
```
MKiperszmid, line 127: What I like to do, is when I have libraries tied to the same version (ex: Hilt has like 5 libraries, all tied to 2.44), i like creating a variable for the version

So in the future, I can just change it in one place, and it gets changed everywhere

app/src/main/java/com/vero/tasky/MainActivity.kt
```diff
@@ -24,7 +24,7 @@ import dagger.hilt.android.AndroidEntryPoint
 @AndroidEntryPoint
 class MainActivity : ComponentActivity() {
 
-    private val viewModel: MainViewModel by viewModels()
+    val viewModel: MainViewModel by viewModels()
```
MKiperszmid, line 27: Why was this made public?

MKiperszmid (approved):

app/src/androidTest/java/com/vero/tasky/auth/presentation/LoginE2E.kt
```diff
@@ -0,0 +1,152 @@
...
+                        RootNavigationTest(
+                            navController = rememberNavController(),
+                            isLoggedIn = composeRule.activity.viewModel.state.collectAsState().value.isLoggedIn,
+                            modifier = Modifier.padding(paddingValues),
+                            loginViewModel = loginViewModel
+                        )
+                    }
+                }
+            }
+        }
```
philipplackner, line 116: For end-to-end tests, I'd not set the content manually. They're intended to test your app as it is in production (just boundaries like API are replaced). For you that means in particular:
1. Don't instantiate your ViewModel here manually. Hilt does that for you.
2. Make Hilt provide a fake API (either as fake or by using MockWebServer) and leave the rest as it is. Use the real repository implementation with the fake API.
3. If you use setContent, then just use it to show the root composable and don't add custom composables like here. They can contain bugs that only occur for your test case, since they don't have anything to do with your production code.

app/src/androidTest/java/com/vero/tasky/core/presentation/navigation/RootNavigationTest.kt
```diff
@@ -0,0 +1,244 @@
...
+    isLoggedIn: Boolean,
+    modifier: Modifier = Modifier,
+    loginViewModel: LoginViewModel? = null
+) {
+
+    NavHost(
+        navController = navController,
+        startDestination = if (isLoggedIn) Screens.Agenda.route else Screens.Login.route,
+        modifier = modifier
+    ) {
```
philipplackner, line 55: Same here, use the NavHost from your production code instead and don't recreate it. This is very error prone and won't catch issues in your real NavHost

philipplackner (approved):

app/src/main/java/com/vero/tasky/MainActivity.kt
AndreVero: I need viewModel to collect isLoggedIn State

MKiperszmid: Where?

philipplackner: But you only use this ViewModel reference in the activity? 🤔

PR #61, opened 2023-07-20
1.0.0 fixes

AndreVero:


app/src/androidTest/java/com/vero/tasky/auth/presentation/LoginE2E.kt
```diff
@@ -104,12 +74,42 @@ class LoginE2E {
...
+                                    onNewAgendaItemClick = { agendaItemType ->
+                                        val route = when (agendaItemType) {
+                                            AgendaItemType.REMINDER -> Screens.Reminder.route
+                                            AgendaItemType.TASK -> Screens.Task.route
+                                            else -> Screens.Event.route
+                                        }
+                                        navController.navigate(route)
+                                    }
+                                )
+                            }
```
philipplackner, line 111: You still implement a copy of the production logic, which is a no go for tests. If your production logic changes, the test logic will remain and keep on passing, even if there's a bug in the production code.
Instead, just include your root composable here which contains the whole nav host:
```
// Production code
@Composable
fun AppRoot(navController: NavHostController) {
    // Your navigation logic here
}

------------------

// Test code
composeRule.setContent {
    val navController = rememberNavController()
    AppRoot(navController)
}
```
You can also take a look at `TestNavController` for more control.

philipplackner (approved):

PR #62, opened 2023-07-22
1.0.0 fixes

AndreVero:


app/src/main/java/com/vero/tasky/agenda/presentation/agenda/AgendaViewModel.kt
```diff
@@ -71,6 +72,13 @@ class AgendaViewModel @Inject constructor(
                     )
                 }
             }
+            AgendaEvent.RefreshAgenda -> {
+                viewModelScope.launch {
+                    updateState(newState = state.copy(isRefreshing = true))
```
MKiperszmid, line 77: The updateState method was used to save the state in case of process death, right?

Personally, I don't save the loading state to survive process death
Since if the user leaves to process death just before line 79, then they will be stuck on a loading screen

MKiperszmid (approved):

app/src/main/java/com/vero/tasky/core/presentation/navigation/AppRoot.kt
```diff
@@ -0,0 +1,57 @@
...
+                        backgroundColor = Color.Red,
+                        snackbarData = data,
+                        contentColor = Color.White,
+                        modifier = Modifier.testTag("snackbar")
+                    )
+                }
+            },
+        ) { paddingValues ->
+            RootNavigation(
+                navController = rememberNavController(),
```
philipplackner, line 51: This nav controller should be passed as a parameter to `AppRoot`, since otherwise you can't control the navigation for test cases. Imagine you'd want to test some interaction on event detail screen and photo detail screen, then you would need to initially navigate to the event detail screen which doesn't work if you don't have the nav controller reference in tests

philipplackner (approved):

